PR #56, opened 2014-12-02
Dynamic paths for nscp.spec

palli:
Just a fly-by commit, since I am not in front of a machine that
can build this rpm package.

Attempting to make destination file in rpm package more dynamic.

This fixes an issue where nsclient gets installed in /usr/lib/
instead of /usr/lib64/ on 64 linux machines.

Please test and give feedback.


mickem:
This has been implemented manually...

PR #108, opened 2015-02-13
Added external script for testing logged in windows users

Vitexus:
This is http://www.nsclient.org/nscp/attachment/wiki/CheckExternalScripts/check_users_loggedin.vbs updated for vb NagiosPlugin class usage.


mickem:
Thanks

Issue #151, opened 2015-05-19
Allow for certificate pinning in NRPEServer


arathai:
Hi,

imho it would be very desirable to allow for certificate pinning in NRPEServer.

What I'd wish for would be a configuration-directive like `trusted_client_cert_file` which takes a path to a certificate in PEM-Format that is trusted and has to be presented by the client. The directive **must be** able to occur multiple times in order to allow two (or more) certificates to be trusted, in order to allow smooth transition in case the certificate on the monitoring Server is changed.
> * create new certificate
> * push certificate to servers and config
> * switch certificate on monitoring-host
> * remove old certificate from servers and config

I'd do a pull-request but as of today I am unable to get the source to build on windows following the [Building NSClient++](http://docs.nsclient.org/manual/build.html) Guide. :/

Best Regards,
Alex

Ps.: This is a followup to [Certificate pinning in NSClient++ (NRPEServer)](https://forums.nsclient.org/t/certificate-pinning-in-nsclient-nrpeserver/3661).

mickem:
Hmm, not sure I understand... Would this not be the ca option in combination with trust?

mgaugusch:
Hi,
I also stumbled across this problem. The problem is, that _any_ Certificate signed by the same CA can be used. So, if a server gets broken into, the attacker could use the server certificate (signed by the common CA) as client certificate to contact all other monitored machines. 
The only workaround would be to use a _different CA_ for each machine, but that's not how it should be done :-)
I would not go as far as storing the full client cert PEM on the server, but at least I would suggest to specify a list of valid subject DNs, that the server accepts. This also allows for renewal of the client cert, without having to update all servers. Of course, if the client cert is compromised, one has to use a different DN then.

mickem:
Already supported in modern versions. The relevant settings under [/settings/NRPE/server] are:

```
[/settings/NRPE/server]
use ssl = true
verify mode = peer-cert
ca = ${certificate-path}/ca.pem
certificate = ${certificate-path}/server.crt
certificate key = ${certificate-path}/server.key
```
Setting verify mode = peer-cert makes the server require, and validate, a client certificate. To restrict to a specific client (true pinning):

1. Use a private CA whose only signed leaf is the monitoring host's cert, and put only that CA in ca = .... Roll-over by issuing a new leaf signed by the same CA, push to clients, switch the monitoring host, then optionally revoke the old one via CRL (crl = ...).
[...]

PR #163, opened 2015-07-02
Using full cert-chain and certificate pinning

### LINKED ISSUE #150: Call to context.use_certificate_chain_file instead of context.use_certificate_file in NRPEServer
Hi,

in socket_helpers.cpp there's a call to `context.use_certificate_file`, if it were `context.use_certificate_chain_file` the Server would be able supply the whole chain and the client wouldn't need to know the intermediate certificates.

It's also the [recommended method](https://www.openssl.org/docs/ssl/SSL_CTX_use_certificate_chain_file.html#NOTES) in the underlying openssl:
>SSL_CTX_use_certificate_chain_file() should be used instead of the SSL_CTX_use_certificate_file() function in order to allow the use of complete certificate chains even when no trusted CA storage is used or when the CA issuing the certificate shall not be added to the trusted CA storage.

See [context.use_certificate_file](http://www.boost.org/doc/libs/1_53_0/doc/html/boost_asio/reference/ssl__context/use_certificate_file/overload2.html) and [context.use_certificate_chain_file](http://www.boost.org/doc/libs/1_53_0/doc/html/boost_asio/reference/ssl__context/use_certificate_chain_file/overload2.html)

Best Regards,
Alex

Ps.: This is a followup to [Certificate pinning in NSClient++ (NRPEServer)](https://forums.nsclient.org/t/certificate-pinning-in-nsclient-nrpeserver/3661).

arathai:
This fixes #150
Tested and working with `openssl s_client` and modified `check_nrpe` under linux.
Note that this causes only PEM-encoded certificates to be accepted which are the default anyway and there's nothing wrong with converting ASN1 certificates using
`openssl x509 -in infile.der -inform DER -out outfile.pem -outform PEM`
Pull Request #162 was for the wrong branch, sorry

### DIFF
--- a/build/python/fetchdeps.py
+++ b/build/python/fetchdeps.py
@@ -225,7 +225,7 @@ sources['lua'] = source('lua-5.1.5.tar.gz', 'http://www.lua.org/ftp/lua-5.1.5.ta
 
 sources['boost'] = source('boost_1_56_0.zip', 'http://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.zip/download', '15b9719a92cd2a80170d54cfb9c3990a56193c56')
 
-sources['openssl'] = source('openssl-1.0.1j.tar.gz', 'https://www.openssl.org/source/openssl-1.0.1j.tar.gz', 'cff86857507624f0ad42d922bb6f77c4f1c2b819')
+sources['openssl'] = source('openssl-1.0.2c.tar.gz', 'https://www.openssl.org/source/openssl-1.0.2c.tar.gz', '6e4a5e91159eb32383296c7c83ac0e59b83a0a44')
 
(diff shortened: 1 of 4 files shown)


arathai:
As you can see I'm farily new to using github... I should have made a branch.
Nevertheless, this now also fixes #151.

New config directive in `[/settings/NRPE/server]`
`trusted client cert`, allows for comma seperated list of path to cert-file

arathai:
/home/travis/build/mickem/nscp/include/socket/socket_helpers.hpp:208:23: error: 
>-- Boost version: 1.46.1
>‘boost::asio::ssl::verify_context’ has not been declared

Well that's true, I used Boost 1.56 like it's stated in fetchdeps.py.
Can I influence the travis build system somehow?

PR #246, opened 2016-02-17
response code of the queryresponse will be returned instead of fix value

florianborgo:
i wanted to use the check_nrpe client on RHEL7 using Icinga2. I've noticed that the response code ($0) of check_nrpe is always 99. In my opinion the return code should be highest return code of all responses.


mickem:
Implemented this slightly differently, thanks for the PR though!
The main difference is that the API allows multiple return packages and thus "only picking the first one" could be a theoretical issue.

PR #319, opened 2016-08-28
external script process stuck

kain64:
process produce output, but stack on exit. need check this status before exit
example:
[image from server](https://snag.gy/M9cx3Y.jpg)


mickem:
Do you have a "testcase" for when this failes?
Sounds very interesting as this is an which have been reported many times...

kain64:
no, I did  not found scenario. reproduced sometimes without specific reasons. for example: we have 50 same virtual pcs, reproduced ~3-2 times. I think need to force external process termination

mickem:
This has been implemented "manually"

PR #405, opened 2017-04-10
Update to new OP5 logo

axolsson:


axolsson:
@mickem A quick fix with the new OP5 branding :)

PR #604, opened 2019-02-13
correct spelling

infotek:
spelling "fond" -> "find"


mickem:
Sorry for the late reply, this is a generated file, but I have created a change in the upstream file so this will be fixed in the next release

PR #640, opened 2019-10-30
Add info to enable module first

Napsty:
This additional small note in the documentation should hopefully be helpful for users falling on their nose the first time they want to use `check_wmi` (like me). The additional info clearly states how to enable the module in order to enable the `check_wmi` command.
This should help avoid future questions as in https://forums.nsclient.org/t/unknown-command-s-check-wmi/4604


mickem:
Sorry for the late reply, this is a generated file so I have created a another PR with a change which updates the script so this will be fixed in the next release.

PR #771, opened 2022-06-20
Update external_scripts.md

TokenRing:
The numbering system for the exit codes is unclear here; you should consider changing the list from 1 through 4, to 0 through 3 - if people were to use 1 through 4, they would end up with their scripts reporting the wrong exit codes!


mickem:
Sorry for the late reply, this was also fixed in another PR which have been merged so this will be fixed in the next release

PR #801, opened 2023-06-23
Debian Dockerfile fixes

dropthemic:
Fixed the Dockerfile for Debian 9 to use archive.debian.org for packages.

Updated python packages to use specific versions as using packages with newest versions are broken because python 2.7 is deprecated. Tested on ARM64 and X64 platforms.


mickem:
Closing as we went in a different direction, sorry...

PR #834, opened 2024-08-14
Update build.md - fix windows instructions format

joba-1:


mickem:
Thanks for your contribution, I have in recent versions removed the build instructions from the docs as it was outdated and we have GitHub actions for building.

PR #1037, opened 2025-08-26
Fix links in docs reference page

The-Rockers:
Hyperlinks in the Documentation's "Reference" page point to nsclient.org/docs/**reference/reference**/[type name]/[module name] which results in 404. Fixed the offending URLs by removing the extra "reference/".


mickem:
Thank you!

PR #1186, opened 2026-03-12
Possible fix for installer wiping config

mickem:


Fantu:
@mickem Still reproduces on a 0.12.6 → 0.13.1 in-place upgrade: `nsclient.ini` is wiped and not recreated.

I tried to have Claude analyze it and he wrote me this (but I haven't checked):

The "Possible fix" (9b51bfbd) doesn't work — and likely makes it worse. The `PurgeConfig` component has `<RemoveFile Name="*.ini" On="uninstall"/>`, which also matches `nsclient.ini`. During an upgrade the guard `NOT OLDERVERSIONBEINGUPGRADED` evaluates **false**, which puts the component in the *absent* state — and an `On="uninstall"` RemoveFile fires exactly when a component goes absent. With `RemoveExistingProducts` scheduled `After InstallExecute` (late), the purge runs *after* BackupConfig/ExecWriteConfig has restored the config, so it deletes the just-restored file. The `NeverOverwrite` on the `nsclient.ini` component then prevents a fresh copy, so nothing is left.

Suggested fix: drop the blanket `Name="*.ini"` removal (it's too broad — it catches the user config). Purge specific files only, and never `nsclient.ini`.

mickem:
That is an interesting suggestion. I will give it a spin. That said it is impossible (I think) to fix old uninstall sequences since they were shipped in the old version so would have to be fixed in the old version already installed.

But all this is messy and fundamentally wrong... But your idea seem like a very good solution: Thanks...

PR #1280, opened 2026-05-22
Update build configurations to use libzip instead of Miniz and remove…

mickem:
… TinyXML2 download on linux


Fantu:
Thanks for this.
I was about to see if an LLM could make those changes without me having to spend too much time reviewing and correcting it.
I tried to tell it to check these commit and wrote that spotted this (I have not checked these but for today I'll probably only test the build on Debian unstable with system packages):

### 1. `FindTinyXML2.cmake` — the pkg-config fallback can never link (latent build break)

`FindTinyXML2.cmake` resolves tinyxml2 via three paths:

1. `find_package(tinyxml2 CONFIG)` → defines the `tinyxml2::tinyxml2` target
2. pkg-config → sets `TINYXML2_FOUND` + `TINYXML2_INCLUDE_DIR` + `TINYXML2_LIBRARIES`,
   **but creates no target**
3. vendored source → `TINYXML2_INCLUDE_DIR` points at the source tree, `tinyxml2.cpp`
   gets compiled in-tree

But the only consumer, `modules/NRDPClient/CMakeLists.txt`, decides what to do
purely from the target:

```cmake
if(TARGET tinyxml2::tinyxml2)
    set(TINYXML2_SRCS "")
    set(TINYXML2_EXTRA_LIBS tinyxml2::tinyxml2)
else()
    set(TINYXML2_SRCS ${TINYXML2_INCLUDE_DIR}/tinyxml2.cpp)   # <-- path 2 lands here
    set(TINYXML2_EXTRA_LIBS "")
endif()
```

So when tinyxml2 is found via pkg-config (path 2 — exactly the "older distro without
the CMake config" case the fallback was meant for), NRDPClient takes the else
branch and:

- tries to compile ${TINYXML2_INCLUDE_DIR}/tinyxml2.cpp, i.e. /usr/include/tinyxml2.cpp,
which does not exist (the -dev package ships only the header + the .so) → configure/build error, and
- never links TINYXML2_LIBRARIES (the -ltinyxml2 from pkg-config), because the only
thing it adds to the link line is TINYXML2_EXTRA_LIBS, which is empty in this path.

TINYXML2_LIBRARIES is set in all three paths but never read by the consumer. On
current Debian/Ubuntu/Fedora path 1 always wins, so the bug is latent today — but
the pkg-config fallback as written cannot work.

Two ways to fix:

- have the pkg-config branch also create an imported target
(add_library(tinyxml2::tinyxml2 INTERFACE IMPORTED) + set its
INTERFACE_INCLUDE_DIRECTORIES/INTERFACE_LINK_LIBRARIES), so the TARGET ...
check stays valid in every path; or
- have the consumer key off TINYXML2_FOUND and use TINYXML2_LIBRARIES plus a
TINYXML2_SOURCE variable (empty unless the vendored path was taken), keeping
the find-module API uniform across the three paths.

2. libzip port — metadata buffer leak (now trivially fixable) + fragile cross-platform ownership

The libzip rewrite of zip_plugin.cpp is functionally fine, but:

- read_metadata() calls extract_file_to_heap() and never frees the returned
buffer — a pre-existing leak, now reproduced with malloc. On the libzip path it
is a plain malloc, so a std::free(p) right after read_metadata(std::string(p))
would close it.
- The ownership contract now differs by platform: the Windows buffer must be released
with mz_free(), the Linux one with free(). If the leak is later "fixed" with a
bare free(), it'll be wrong on Windows. Cleanest would be to have
extract_file_to_heap return a std::string/std::vector<char> (RAII), removing
the manual free and the allocator mismatch entirely.
- Minor, while in there: extract_file_to_heap treats malloc(0) returning nullptr
as a failure (would break on a legitimately empty entry), and st.size is used
without checking st.valid & ZIP_STAT_SIZE.

Fantu:
I tried to build on Debian unstable without some parts (webserver, docs, tests and check_nsclient) and seems detect correctly tinyxml2 and libzip from system packages but build failed multiple times for parts missed.
I tried to use Claude Code to create a patch to make certain parts truly optional, and it worked at first, but then it failed with the Boost library. Made a small change there as well, but it failed when using the Debian unstable version, throwing dozens of errors that seemed to be related to missing Boost functions.
For today I stopped the test.

Here the branch I used for tests and the patch created with claude:
https://github.com/Fantu/nscp/commits/agent-only-build/
the latest commit was only for continue some fast test and force to use system boost but the actual Debian unstable one seems "too new"
about https://github.com/Fantu/nscp/commit/d4514b912473131262e803ed68c9e58e27ee86dc I think can be useful if you want to take it or give a look.
I only ran a few checks and tests for less than two hours because I was already tired this evening, so I didn't thoroughly review the PR that I had Claude review; I've pasted the section with the errors he found. I also didn't thoroughly review the patch about the CMake options, but it looked fine at a quick glance.

mickem:
This is just a quick spike (wanted to get a PR out so we dont double work on this) so code has yet to be finalized/reviewed and validated I was mainly curious if it would even build.

I am currently porting all integration tests (the ones which test various protocols) so they also runs on linux which makes it easier to ensure that dependencies actually work i.e. tinyxml is used by NRDP (test by nrdp integration test) and cryppto++ is used by NSCA (tested by nsca integration test). So that will be the next step (which will be out in a bit).
After that I will review the code and ensure things work (and we have good test coverage).
I also noted that some of the FOUND_ options are badly named as well so I will revisit the came parts as well...

mickem:
regarding the unstable build seems to me that many errors stem from google-test so seems like something with google-test or possibly google-test versus boost?

Fantu:
Yesterday I wanted to run some quick tests to see if I could compile it using system packages and with a few things disabled, so I could start working on the packaging if I'll have time.
Unfortunately, there were more prerequisites to fix than expected, even though I tried to get some quick patches done on Claude Code.
I’m still pretty new to using LLMs and have perhaps just learned the basics of how to use them to make them somewhat useful (though it seems highly unlikely to me that they could be developed independently by any novice, as many would have us believe).
I tried using it during the major task of Debian packaging a new major version of Cinnamon and all its components; it took me 30–40 hours in total over the course of a couple of months to do what I would have done completely on my own, excluding the full review of dependencies and copyrights, but I did also that as well.
In general, with the packaging had several errors or outdated things, so while it was useful for d/copyright reviews, which would have been a massive task to do and check manually (despite the use of some specialized tools), and also for check dependencies in the code (to see which ones are not used anymore to be removed, and which ones were missing but included by other dependencies), so I suppose it saved me at least twenty hours on that part.
Maybe I'll have him create the initial packaging design for NSCP (excluding a few parts) and see if I won't have to spend too much time guiding him and correcting the result.
Meanwhile, when it comes to translations, LLMs are driving me crazy and wasting a ton of my time. Ever since translators started using LLMs, the quality has gone downhill, and by a lot. Too many parts get cut or altered, and often the meaning gets lost or even reversed.

mickem:
I would say LLMs have to be used with caution... great tools but they are not (as people will make claim) actually intelligent. So they make "something" which is "likely" not correct and not whats required. I tend to use them a lot (and they are efficient) but frequently have to convince it to do what we want and not what it thinks is the correct answer. They aso frequently take very long time an require a lot of round-trippping making them sometimes slower then doing it by hand :)

Fantu:
I know that LLMs should be used with caution, and I’ve already seen firsthand that in many cases, when it comes to topics we have fairly knowledgeable about, may just end up wasting more time.
I’ve also realized that they aren’t true "AI" (and I suppose they never will be), but rather they search for matches in the available data and provide probable answers based on the data they have (which could be incomplete, outdated, or incorrect; which is why data is the most important thing to invest time in to improve results) and specific settings.

As for this PR, I first rebased the initial commits of today and tried the build again.
I don't use the gtest part (with tests disabled), so I don't know if it compiles correctly, and I assume that's where your test got stuck for errors on it.
Regarding Boost, I now see:
`Could NOT find boost_system (missing: boost_system_DIR)`
I'll try to take a closer look.

Fantu:
For the test I used https://github.com/Fantu/nscp/commits/agent-only-build/ and:
```
cmake -S . -B build-test -G Ninja \
    -DBUILD_WEBSERVER=OFF -DBUILD_DOCS=OFF -DBUILD_TESTING=OFF \
    -DNSCP_BOOST_PYTHON_VERSION=python313 \
    -DCHECK_NSCLIENT_MISSING=ON \
    -DCMAKE_BUILD_TYPE=Release
cd build-test && ninja -k 0
```
[nscp-sid-build.log](https://github.com/user-attachments/files/28175002/nscp-sid-build.log)
[nscp-sid-configure.log](https://github.com/user-attachments/files/28175003/nscp-sid-configure.log)

On configure I saw warning on iconv that seems strange (probably something wrong) but don't seems blocking, not checked more.

On build about boost seems that nscp uses deprecated functions, some even from 2017 and then removed in boost 1.87 :(
To save time, I tried having claude search for all the features missing from the build, checking whether they were deprecated or removed and when.
Here the result:

### Boost.Asio APIs used by NSCP that broke on Boost 1.90 (Debian unstable)

All errors come from the network headers (`include/net/socket/*.hpp`, `include/net/http/client.hpp`). Below are the root-cause APIs only (the many `'... was not declared in this scope'` errors are just downstream consequences). NSCP does **not** define `BOOST_ASIO_NO_DEPRECATED`, so this is the default Boost behaviour, not an opt-out.

| API used by NSCP | Status in Boost 1.90 | Deprecated | Removed |
|---|---|---|---|
| `boost::asio::io_service` | removed (use `io_context`) | Boost 1.66 (Dec 2017) | Boost 1.87 (Dec 2024) |
| `boost::asio::mutable_buffers_1` | removed (use `mutable_buffer`) | Boost 1.66 (2017) | Boost 1.87 (2024) |
| `boost::asio::ip::tcp::resolver::query` | removed (use `resolve(host, service)`) | Boost 1.66 (2017) | Boost 1.87 (2024) |
| `boost::asio::ip::tcp::resolver::iterator` | removed (use `resolver::results_type` / range) | Boost 1.66 (2017) | Boost 1.87 (2024) |
| `boost::asio::ip::address_v6::is_v4_compatible()` | removed (IPv4-compatible addresses dropped) | Boost 1.66 (2017) | Boost 1.87 (2024) |
| `boost::asio::ip::address_v6::to_v4()` | removed (IPv4-compatible addresses dropped) | Boost 1.66 (2017) | Boost 1.87 (2024) |
| `boost::asio::ssl::rfc2818_verification` | removed (use `host_name_verification`) | Boost 1.73 (2020) | Boost 1.87 (2024) |
| `boost::asio::deadline_timer` | **present but no longer pulled in by** `<boost/asio.hpp>` | Boost 1.87 (2024) | **not removed** (prefer `steady_timer`, or `#include <boost/asio/deadline_timer.hpp>` + Boost.Date_Time) |
| `boost::posix_time::time_duration` | exists (Boost.Date_Time, not Asio) | — | **not removed**: it used to be included transitively via `deadline_timer`; now needs an explicit `#include` |

**Notes**

- Every "removed" API above was confirmed **absent from the installed Boost 1.90 headers** (Asio 1.38); the removal happened in the **Boost 1.87 / Asio 1.33 cleanup (Dec 2024)**.
- The first six were deprecated back in **Boost 1.66 (2017)** — roughly a 7-year deprecation window. This is the same wall many C++ projects hit at Boost 1.87 (e.g. monero, OpenROAD), not a Boost regression.
- The only two that are *not* removed (`deadline_timer`, `posix_time::time_duration`) fail only because `<boost/asio.hpp>` stopped including them; the clean fix is to migrate to `steady_timer`.
- `get_io_service()` shows up in the errors but **still exists** (returns `io_context&`); it fails only because a `deadline_timer` sits next to it.

mickem:
Hmm, nice find.. Most of the asio stuff is "ancient" so likely due for an overhaul. 
That said a challenge is that boost is no longer very backwards compatible as they dropped support for older windows (which is why I am using the version I use).  I have plans (not realized yet) to handle this by splitting the "legacy windows version" from the "normal version" having them build with different compilers and different lib versions (now the only real difference is linkage). 
But that is likely a month or two off (pan to change the installer first).
So might be a bit more work then to get this to build natively...

If you want to take a stab at boost asio I would break that into a separate feature(s)... making small changes there which are self contained (easy to review/merge)... 

I am exploring (another spike) replacing mongoose with boost-beast which will use the newer overlays for asio but havent started to do any real heavy lifting there yet (maybe tomorrow) that might give me some understanding of the new boost asio and how hard it is to migrate/support both...

mickem:
Hmm...

Windows is are currently at boost 1.86 so if they were deprecated before then I guess that might be less of an issue actually? Feel free to have your AI to spike it and see if the integration test in this PR runs... If they do we are likely pretty safe... The integration test (requires docker) will run protocol test with various things so not stress test and not security but at least functional tests...

Fantu:
I didn't have Claude try to create a patch because these seem like delicate parts, and testing the build isn't enough. If he could also run tests to find regressions and go through a “loop” of trial-and-error fixes, that might give him a chance to get something right (if he doesn't run out of tokens first).
I’ll maybe try again another time; right now I was curious to see what Debian package he’d generate if I guided him only a little.

mickem:
FIY: I forced pushed this branch (will merge it after builds are done) so beware if you update... 

Also I took a stab at boost asio migration and that was pretty much search/replace (no idea if it compiles on unstable though, but at least there might be fewer compilation errors): https://github.com/mickem/nscp/pull/1281. only last commit as the PR currently includes this PR as well, will fix that once this is merged....

PS: if you're not handy with git easiest (in my humble oppinion) way to migrate to main after this is merged is to create a new branch and cherry pick your commits there...

Fantu:
You did a huge work.
From a fast look on zip changes after the initial check of https://github.com/mickem/nscp/pull/1280#issuecomment-4521774858 seems that both issue is still present.
I'll retry the build with new main+PR1281

mickem:
You are right as I was reviewing the code I decided it was getting messy and wanted to move the messiness out but thought the PR was big already so deferred it to this PR instead: https://github.com/mickem/nscp/pull/1284

PR #1281, opened 2026-05-23
Migrated deprecated asio apis

mickem:


Fantu:
I did a fast test with this on https://github.com/Fantu/nscp/commits/agent-only-build2/ and still fails but seems because is a WIP.

I not have time to check the log, I simply did a rebase and cherry-pick of this, launched the new test and used claude to check the result, I copy here, if can be useful:

**Result:** CMake configure is clean; the build now gets *much* further than before the Asio
work (the old ~259 Asio errors are gone), but it still fails. `ninja` stops at ~67/568 with
37 errors, all in `include/net/socket/socket_helpers.hpp` and `include/net/socket/client.hpp`.

## Root cause

`a15d04dd` adds `-DBOOST_ASIO_NO_DEPRECATED` to `NSCP_GLOBAL_DEFINES`. In Boost ≥ 1.87
`boost::asio::deadline_timer` (and the rest of the Boost.Date_Time time-traits support) lives
behind the `!defined(BOOST_ASIO_NO_DEPRECATED)` guard, so turning that macro on **removes
`boost::asio::deadline_timer` entirely**. The commit converted most of the net stack
(`io_service`→`io_context`, resolver, buffers, …) but the **timer machinery was not finished**:
several files still declare `(boost::)asio::deadline_timer` and call the deprecated
`expires_from_now(boost::posix_time::…)`. Each of those is now a hard error.

Errors actually hit by the build (the first two files reached):

- `include/net/socket/client.hpp:36` — `connection<>::timer_` is `boost::asio::deadline_timer`
  (timeout member is `boost::posix_time::time_duration`, line 37); used at lines 68/74.
- `include/net/socket/socket_helpers.hpp` — `io::timed_writer` (254), `io::timed_reader` (329)
  and the free functions `write_with_timeout` (304) / `read_with_timeout` (378), all
  `deadline_timer` + `boost::posix_time::time_duration` params.

## Complete list of remaining `deadline_timer` (9 files)

The build stopped early, but the same failure is latent in every file below — all will fail
identically once compiled with `BOOST_ASIO_NO_DEPRECATED`:

| file | line | note |
|---|---|---|
| `include/net/socket/client.hpp` | 36 | **confirmed** — `connection::timer_` |
| `include/net/socket/socket_helpers.hpp` | 254, 304, 329, 378 | **confirmed** — timed_writer/timed_reader + free fns |
| `include/net/socket/connection.hpp` | 225 | `timer_`, `expires_from_now(posix_time::seconds)` @122 |
| `include/net/pinger.hpp` | 134 | `timer_`; uses `expires_at(posix_time::ptime)` @90 (absolute time) |
| `modules/CheckNet/check_dns.cpp` | 64 | `expires_from_now(posix_time::milliseconds)` @72 |
| `modules/CheckNet/check_tcp.cpp` | 65 | `expires_from_now(posix_time::milliseconds)` @79,129 |
| `modules/CheckNet/check_ntp_offset.cpp` | 71 | `expires_from_now(posix_time::milliseconds)` @108 |
| `modules/NSCANgClient/nsca_ng_client.cpp` | 171 | `expires_from_now(posix_time::seconds)` @255 |
| `modules/SMTPClient/smtp.cpp` | 127 | `asio::deadline_timer` alias; `expires_from_now(posix_time::seconds)` @128 |

(`modules/CollectdClient/collectd_client.hpp:55` has a commented-out `deadline_timer` — no action.)

## Suggested migration

- `(boost::)asio::deadline_timer` → `boost::asio::steady_timer` (relative timeouts) or
  `boost::asio::system_timer` where an absolute wall-clock instant is needed (`pinger.hpp`,
  which uses `expires_at(posix_time::ptime)`).
- `expires_from_now(boost::posix_time::seconds/milliseconds(n))` → `expires_after(std::chrono::seconds/milliseconds(n))`.
- Durations crossing API boundaries should become `std::chrono` types. This ripples to callers
  that currently pass `boost::posix_time::time_duration` / `posix_time::seconds`, e.g.:
  - `connection`/`tcp_connection`/`ssl_connection` ctors (`client.hpp:45,228,264`) and the
    proxy at `client.hpp:388` (`posix_time::seconds(info_.timeout)`);
  - `include/net/check_nt/client/socket.hpp` (`send`/`recv`/`read_with_timeout`/`write_with_timeout`
    take `boost::posix_time::seconds` and forward into `socketHelpers::io::*_with_timeout`).

Unrelated note: the `posix_time` usages combined with `boost::get_system_time()` for
Boost.Thread timed mutex/locks (CheckDisk, CheckEventLog, CheckSystem, …) are **not** affected —
those are Boost.Date_Time, not Asio.

The agent-only CMake switches and the `CMP0167 NEW` (BoostConfig) change are independent of this
and configure/compile fine; the only remaining blocker for an agent-only build on Boost 1.90 is
finishing the timer migration above.

mickem:
Feel free to run again, and see if there are any other issues. The installer test fail due to a change on main so this is "kind of ready for merge"...

Fantu:
boost major of issues seems solved
I had Claude make the necessary changes and also check the minimum required Boost version based on the functions used, which I think will be helpful:
https://github.com/Fantu/nscp/commit/a71add8f26880afa8c44b85cb35855ea2010aac1
https://github.com/Fantu/nscp/commit/25172b50c539c7aba8586fb1b5c8fadaeb19cdc3

From what I've seen, I don't think CMake version 3.10 is the real min. version anymore. I'll ask Claude to check all the functions we're using and see what the actual minimum version requirement is.
As for CMake in projects I've worked on, a few things really ate up a lot of my time. To support very old versions, we were forced to do extra work, some of which wasn't ideal, so maybe raising the minimum version requirement could help improve things.

Fantu:
Result:
## `cmake_minimum_required(VERSION 3.10)` is lower than what actually builds

Every `CMakeLists.txt` declares `cmake_minimum_required(VERSION 3.10)`, but the
project cannot be configured with CMake 3.10/3.11: `dependencies.cmake` calls
`find_package(Python3 COMPONENTS Interpreter Development)` unconditionally, and
the `FindPython3` module only exists from CMake 3.12. On older CMake, Python3
isn't found and the build hits the `message(FATAL_ERROR)` for missing required
dependencies. So the real floor is 3.12. (The default, non-agent build also
uses `FetchContent_MakeAvailable` for googletest, which needs 3.14.)

Not urgent, but the declared minimum could be bumped to 3.12 (or 3.14) to match
reality. Both are still above the 3.10 threshold that recent CMake deprecates,
so this stays compatible with CMake 4.x.

mickem:
Yeah, regarding cmake: I saw that as well and I have created an issue to track this (and various other) cmake issues here: https://github.com/mickem/nscp/issues/1289

mickem:
Regarding your commits: 
* https://github.com/Fantu/nscp/commit/25172b50c539c7aba8586fb1b5c8fadaeb19cdc3 create a PR and I will merge seems reasonable to include correct files and something I'm always fixing when I see it.
* https://github.com/Fantu/nscp/commit/a71add8f26880afa8c44b85cb35855ea2010aac1 We cannot hardcode the boost version but Feel free to create an IF(not windows) remove system and (else) include it and create a PR and let see f that flys, I guess on older os:es that might be an issue but not sure what dists ship versions prior to that and also not sure if we actually link system. Might be worth looking into if we use any features from system that requires linkage.

Fantu:
> * [Fantu@a71add8](https://github.com/Fantu/nscp/commit/a71add8f26880afa8c44b85cb35855ea2010aac1) We cannot hardcode the boost version but Feel free to create an IF(not windows) remove system and (else) include it and create a PR and let see f that flys, I guess on older os:es that might be an issue but not sure what dists ship versions prior to that and also not sure if we actually link system. Might be worth looking into if we use any features from system that requires linkage.

1.75 is the minimum version required and is good to know and check, so it warns you if you try to use an incompatible minor version, and it seems to me that a function that requires that version was already in use before the recent changes.
Regarding setting a minimum version in cmake, I can't remember of any better methods at the moment, but if there are any, use them. The important thing in my opinion is that you know and check the minimum version.

mickem:
Ohh, if it is the minimum version then it makes sense.

I just checked on windows we have 1.86 so this should be fine on windows as well so create a PR with both these changes...

PR #1284, opened 2026-05-24
Moved unzip to a reusable module

mickem:


Fantu:
from a fast look the issue spotted by claude about tinyxml2 and zip seems solved

PR #1290, opened 2026-05-25
Fix remaining boost issues

Fantu:
- add min. boost version required
- remove require of system component that give a "false" module not found, Boost.System has been header-only since Boost 1.69
- add some include to make build works on latest versions where are not declared in other part anymore

more details in commits description

### DIFF
--- a/build/cmake/dependencies.cmake
+++ b/build/cmake/dependencies.cmake
@@ -110,9 +110,8 @@ if(NSCP_WEB_BACKEND STREQUAL "beast")
 endif()
 
 find_package(
-    Boost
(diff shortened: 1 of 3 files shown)


mickem (approved):

mickem:
Would you mind closing and opening a new PR? I have made a change to my pipelines and want to see if they run now?
Sorry, but had a miss configuration in them before...

Fantu:
a push or force push is enough, now remain only you to approve for run workflow (if I remember good because is my first contribute here)

mickem:
Thanks

PR #1291, opened 2026-05-25
Pipelines should run on PRs

mickem:


.github/workflows/build-feature.yml
```diff
@@ -4,8 +4,8 @@ permissions:
   contents: read
 
 on:
-  push:
-    branches-ignore: [mob/**, main]
+  pull_request:
+    branches: [ "main" ]
```
Fantu, line 8: @mickem I think can be useful to don't restrict target to "main", so will test also PR that add commits to existent PR, for example if you work on one big thing in one branch and open a PR related, and a contributor want to do a PR for add a fix commit on it

PR #1298, opened 2026-05-27
Missing dependencies should degrade nicely. Fixed all dependencies no…

mickem:
…w degrade nicely (added test for this under build/docker/*)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+# Build context filter for the Docker images under build/docker/ (and any other
+# `docker build .` run from the repo root). Keeps the context small and the
+# build hermetic — host build trees must never leak into the container build.
+#
+# NOTE: this only excludes the hyphenated out-of-tree build dirs (build-asan,
+# build-cov, …) and cmake-build*; the in-tree source directory `build/`
+# (build/cmake, build/docker, build/python) is intentionally NOT matched.
+.git
+.idea
+.vscode
+cmake-build*
+build-*
+**/node_modules
+rust/*/target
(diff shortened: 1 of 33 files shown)


Fantu:
Thanks for these improvements.
I only gave a fast look on some parts, small things in build.md, cmake >=3.14 and boost >= 1.75, I not searched about effective min. version of other deps.; and the web server is showed as optional, but I didn't saw something that made it possible to disable it but only select the beast or mongoose

mickem:
This PR is only "half" the puzzle, the next PR will add options to disable modules.
That said if you build the web server with beast it does not have the mongoose third party dependency (beast is part of boost). So one option for packaging the web server is to include it as a "monitoring only" component and have a script to download the "web site" so that you dont need npm for building. So Thats something I will support "soon:ish".

I am also working on fixing all paths so they work with prefixes like `/usr/local` as well as adding options to disable modules on build. All this hopefully out in the next few days.

Fantu:
Thanks.
Currently it seems to only support / as a prefix, while major distros use /usr and some users use /usr/local, supporting at least those while respecting FHS in generated paths would be great.
Then there are the public shared libraries that actually appear to be private. Can you confirm that they are private? Or do they need to be exposed and used by other external programs that are not NSCP components?
Regarding the web part, downloading from a script to avoid the hassle of packaging would probably not be accepted or at least discouraged by Debian, so I'll try to add that part to the package if the necessary packages already exist in Debian or if it's not too long to do (I haven't checked that part yet).
I don't know if Claude is able to search correctly. The initial search for the main part made several errors that I then had to verify manually. I realized that to do it correctly, I should instead use Claude Code in the repository, have him search carefully in the code, and also check the actual and current dependencies in the Sid chroot. As for the web packages, however, it seems to me that there's a risk that I'm making mistakes. I should first find out all the various "naming conventions" used and perhaps also have him search the respective teams' package lists to ensure that he generates correct results.

PR #1307, opened 2026-06-03
cmake: allow disabling the web layer via BUILD_MODULE_WEBServer

Fantu:
Upstream always builds the Beast/mongoose HTTP backend library: libs/mongoose-cpp (nscp_mongoose) is added by the unconditional libs/* glob and installed by NSCP_MAKE_LIBRARY, the HTTP backend selector in dependencies.cmake runs and aborts when its mongoose default cannot find the Windows-only vendored source, and the mongoose wrapper test links nscp_mongoose. Disabling only the WEBServer module via the generic per-module flag (-DBUILD_MODULE_WEBServer=OFF) therefore still builds and
installs that library orphaned, still runs the backend selector and still
requires a backend.

Rather than add a new top-level option, reuse the existing per-module flag that LOAD_SECTIONS already declares for every module, so a single -DBUILD_MODULE_WEBServer=OFF turns off the whole web layer in one place (same flag name and spelling as every other module):

- CMakeLists.txt: declare option(BUILD_MODULE_WEBServer ...) up front, before dependencies.cmake and the libs/* glob, so those early consumers
see it (and its ON default). LOAD_SECTIONS' later option() call for the
  same name is then a no-op, and it keeps honouring the flag natively to
  skip the module itself - so modules/WEBServer/module.cmake needs no
  change.
- CMakeLists.txt: skip libs/mongoose-cpp in the libs glob when it is OFF.
- build/cmake/dependencies.cmake: run the HTTP backend selector only when
  it is ON, so a no-web build needs neither the vendored mongoose source nor Beast/OpenSSL and does not abort.
- tests/CMakeLists.txt: gate the mongoose wrapper test, which links nscp_mongoose (no longer built).

With BUILD_MODULE_WEBServer=ON (the default) nothing changes: configure on Debian sid still selects the backend (e.g. beast) and adds the mongoose library as before. With -DBUILD_MODULE_WEBServer=OFF the backend
selector is skipped, no mongoose backend library is built, and the WEBServer module is reported as disabled - a build with no web/HTTP server at all (e.g. a stripped-down agent package).

Assisted-by: Claude Code (Opus 4.8)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,18 @@ include(${BUILD_CMAKE_FOLDER}/GoogleProtoBuf.cmake)
 # functions.cmake so NSCP_CREATE_TEST can honour it.
 option(NSCP_BUILD_TESTS "Build the unit tests (requires Google Test)" ON)
 
+# WEBServer's per-module flag (the generic BUILD_MODULE_<name> option that
+# LOAD_SECTIONS declares for every module) is consumed *early* here — by the
+# HTTP backend selector in dependencies.cmake and by the libs/* glob that adds
+# its Beast/mongoose backend library (libs/mongoose-cpp -> nscp_mongoose) — long
+# before LOAD_SECTIONS would normally declare it. Declare it up front so those
+# early consumers see it (and its ON default); LOAD_SECTIONS' later option()
+# call for it is then a no-op. With -DBUILD_MODULE_WEBServer=OFF this turns off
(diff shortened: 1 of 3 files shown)


mickem:
I have been pondering this and I think the problem here is that we bring in "all libraries" as dependencies (instead of using the actual dependencies). So if we remove that this would likely "sort it self out"...

mickem:
Here is how that looks: https://github.com/mickem/nscp/pull/1308/changes

PR #1308, opened 2026-06-04
Allow building with missing beast/mongoose if web server is disabled.

mickem:


Fantu:
I'm testing now and with -DBUILD_MODULE_WEBServer=OFF and -DNSCP_WEB_BACKEND=mongoose (default) don't build the webserver part but if in future will be -DNSCP_WEB_BACKEND=beast the default with boost and openssl installed will build even if -DBUILD_MODULE_WEBServer=OFF, or I'm wrong?

mickem:
**1**: `nscp_mongoose` should be named `nscp_web_lib` or something along those lines (but renaming it is expensive) so it will be used regardless of `NSCP_WEB_BACKEND` i.e if use boost::beast over mongoose it will still be called nscp_mongoose.

**2**: If `BUILD_MODULE_WEBServer=OFF` there is no need for `nscp_mongoose` so it will not be built (it will still be configured, but never built as there is nothing requiring it).

**3**: CMake will accept a broken `nscp_mongoose` (i.e. missing mongoose or missing openssl or missing beast) unless the module is actually required.

The effect of this is that if you disable the `WEBServer` the value of `NSCP_WEB_BACKEND` does not matter as it will not be built and neither will the various issues like missing mongoose or missing openssl or missing beast.

That said there could be bugs as I have not tested all possible combinations, but thats the intent...

Fantu:
I did a fast test with -DNSCP_WEB_BACKEND=beast and build the part even if BUILD_MODULE_WEBServer=OFF so if in the future the default will change to beast BUILD_MODULE_WEBServer=OFF will not be enough for really disabled the webserver part.

mickem:
Ahh... wait I build the cmake target, while I guess you dont and thus build "all"... 

Let me try to fix that...

mickem:
I did a follow up commit here https://github.com/mickem/nscp/commit/8cccbdd82baa522d1731a27da0d6e63341eab3bb (folded into another PR as build times are so slow and did not want to wait for another one). So with this (main) it should work as expected now I hope (please feel free to verify).

Thanks for reporting this...

PR #1312, opened 2026-06-06
feature: add support for custom install prefix on Linux builds

mickem:


Fantu:
I not checked deeply, but now is better; before it was supporting only "/", now support the more used path, I not tested if it is also okay with different paths.
One thing spotted is the linux-prefix-builds.md is mentioned but not present

PR #1314, opened 2026-06-09
add codespell github workflow

Fantu:
this help showing typo errors in the pull request
show also typo out of PR changes
setted to only warn, so will don't fails the check even if typo are present, thus it does not force to correct any typo, which may not even be introduced by that PR

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,17 @@
+name: Codespell
+on: [pull_request]
+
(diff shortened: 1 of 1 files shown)


Fantu:
@mickem about rust folder that was moved to another repository, is it not better to remove it?

mickem:
sorry, I completely missed this one...

mickem:
Thanks...

Issue #1076, opened 2025-11-25
Question about check_process


Fantu:
Hi, I'm trying to improve check_process output, I did a check for possible memory leak on some windows servers:
`check_process "filter=exe not in ('sqlservr.exe')" "warn=working_set > 3G" "crit=working_set > 5G"`
on the perf data I want to view top 10 or 15 process with more use of ram, I tried but seems is not possible order for ram usage, or I'm wrong?
I also tried to filter for select only process >500mb but failed
`"filter=exe not in ('sqlservr.exe') and working_set > 500m"`
Is possible to filter to exclude some processes for name and also for lower ram usage?

I tried on windows 2019 with nsclient++ 0.10.8

@mickem I found that "[Download the latest version](https://nsclient.org/download)" on the site isn't working, I suppose you need to use for example this link to point the latest release: https://github.com/mickem/nscp/releases/latest

mickem:
Sorry for the late reply but been a busy week...

Anyways...

You can use [filter_perf](https://nsclient.org/docs/reference/check/CheckHelpers/#filter_perf) to post process performance data.

This works by executing a regular check as a sub command and the post processing the result of that check (think of it like a pipe in linux)

In its most basic form you can run `filter_perf command=COMMAND arguments REGULAR ARGUMENTS` in your case:
```
filter_perf command=check_process arguments "filter=exe not in ('sqlservr.exe')" "warn=working_set > 3G" "crit=working_set > 5G"
L        cli WARNING: WARNING: clion64.exe=started
L        cli  Performance data: ' ws_size'=0GB;3;5 ' ws_size'=0GB;3;5 ' ws_size'=0GB;3;5 ' ...
```

This will not do an anything by it self but now you have the option for instance to sort the performance data by adding `sort=normal`:
```
filter_perf sort=normal command=check_process arguments "filter=exe not in ('sqlservr.exe')" "warn=working_set > 3G" "crit=working_set > 5G"
L        cli WARNING: WARNING: clion64.exe=started
L        cli  Performance data: 'clion64.exe ws_size'=3.30851GB;3;5 'Rider.Backend.exe ws_size'=1.80017GB;3;5 'clangd.exe ws_size'=1.4822GB;3;5 'devenv.exe ws_size'=1.14938GB;3;5 ...
```

As you can see we now have the biggest process at the top.

Now this still return a million entries so you can add `limit=5` to only return the first 5 values:
```
[...]

PR #1326, opened 2026-06-29
docs: document filter_perf for top-N processes and clarify check_process --delta

Fantu:
Today I had to add a check similar related to the ram (but for cpu).
I didn't find it from a quick look at the documentation, but I remembered what [mickem wrote a while ago about RAM](https://github.com/mickem/nscp/issues/1076), so I haven't had problems. Other users may have trouble finding the same things, which would risk to wasting time and creating custom scripts.
So I tried to have Claude improve the documentation a bit; it also found something wrong with the description of --delta. Unfortunately, I don't have enough time today to generate updated documentation and check it carefully, so it probably needs improving or fixing.

### DIFF
--- a/docs/samples/CheckSystem_check_process_desc.md
+++ b/docs/samples/CheckSystem_check_process_desc.md
@@ -1,0 +1,38 @@
+#### Showing only the top processes (sorting and limiting)
+
+`check_process` does not sort or limit its output: every matching process is evaluated and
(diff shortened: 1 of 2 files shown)


mickem:
Why did you close this?

Fantu:
I had Claude create it and committed it without checking first, it had already made an error (--delta -> delta=true), I will do it again when I have the commands that work and I will be sure that there are no other errors.

PR #1330, opened 2026-07-01
docs: cross-reference filter_perf from check_process for top-N processes

Fantu:
done again, rebased and removed delta part (delta desc already fixed by mickem)
@mickem I wanted to try checking the regenerated documentation and looking at it locally, but it seems like it doesn't generate the Windows-only parts like that from my Linux workstation.
So I quickly checked the text and it doesn't seem to have any errors (unlike the last time), but I haven't checked the regenerated documentation.


mickem:
Thanks

PR #1331, opened 2026-07-01
improve CLAUDE.md

Fantu:
The first commit improves the text related to sign-off to prevent Claude from making mistakes by using a fixed name of the example, and I've come up with other things to specify to try to avoid further unexpected issues.

It's not a guarantee of avoiding problems given how LLMs are designed. I knew they were probabilistic, and I've also seen how translators have gotten worse since they started using LLMs (who cut and modify sentences). But I'm encountering several cases where they violate simple memory rules, so they're even more unreliable than I thought. However, I think it's worth at least trying to avoid problems. Memory should give them at least a good chance of following these instructions, even if it's not as certain as normal programs (bugs and hardware problems excluded).

In second commit I propose to put also assisted-by in the commit, based on kernel policy that seem me good.

@mickem what do you think?

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,9 +1,17 @@
 # Project conventions
 
 ## Git commits
-- Always add a DCO `Signed-off-by` trailer to every commit (this project requires DCO):
-  `Signed-off-by: Michael Medin <michael@medin.name>`
-  Equivalent to committing with `git commit -s`.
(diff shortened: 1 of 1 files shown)


Fantu:
I think will be good add also CONTRIBUTING.md including that part and also all useful things, especially for manual contributions but could also be useful for those with LLM

PR #1334, opened 2026-07-04
cmake: derive the default python lib name instead of hardcoding versions

Fantu:
On Linux PythonScript resolves NSCP_BOOST_PYTHON_VERSION through a hardcoded if/elseif chain (python39 ... python313) whose only purpose is to define DEFAULT_PYTHON_LIB, the default value of the "python lib" setting (the libpython loaded at runtime). Any Python release newer than the chain aborts configure: with Python 3.14 (now the default python3 in Debian sid, with Boost 1.90 shipping libboost_python314) it fails with 'Unknown python version: python314'.

Replace the chain with a generic match that derives the library name from the component name (python3XX -> libpython3.XX.so). Behaviour is identical for every version the chain listed, future Python releases work without touching this file again, and a component name that does not match the pattern still fails configure with the same error message.

### DIFF
--- a/modules/PythonScript/CMakeLists.txt
+++ b/modules/PythonScript/CMakeLists.txt
@@ -34,30 +34,13 @@ if(WIN32)
         ${NSCP_DEF_PLUGIN_HPP}
     )
 else()
-    if(NSCP_BOOST_PYTHON_VERSION STREQUAL "python313")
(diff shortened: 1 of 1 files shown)


mickem:
Thanks

Fantu:
adding and testing autopkgtest in debian packaging I found another improvement related to this and added in the new commit

PR #1339, opened 2026-07-05
Added check security

mickem:


Fantu:
Thanks for these additions. Regarding the Windows firewall, it would be very helpful to optionally report warnings if it switches to public.
I've had it switch from private/domain to public many times due to router/connection changes that clients don't report, and other reasons. This has led to problems with services being blocked because the rules were only on private and/or domain, not public.

mickem:
> Thanks for these additions. Regarding the Windows firewall, it would be very helpful to optionally report warnings if it switches to public. I've had it switch from private/domain to public many times due to router/connection changes that clients don't report, and other reasons. This has led to problems with services being blocked because the rules were only on private and/or domain, not public.

Added

Fantu:
Thanks, you're adding so much stuff so quickly, it'll take me a while just to read and test what I need^^'

PR #1345, opened 2026-07-05
Fix REUSE third-party attributions and literal SPDX tags

Fantu:
@mickem
One caveat on the new `wixca.dll` annotation: I tagged it `CPL-1.0` for consistency with the WiX UI files already listed in THIRD-PARTY-NOTICES.md, but the DLL was added to the tree in 2014, i.e. the WiX 3.x era, and WiX had switched from CPL-1.0 to MS-RL around 3.5. If you can tell which WiX release it was built from, it may need retagging to `LicenseRef-MS-RL` (and a matching text file), or ideally rebuilding it from a current WiX so the license question goes away.

After this change `reuse lint` has one remaining complaint: `Unused licenses: OFL-1.1, Zlib`. Those two texts only cover binary dependencies (Roboto in the web bundle, TinyXML-2), no tracked file uses them, and the REUSE spec reserves `LICENSES/` for licenses referenced from the tree. I deliberately left that to you; the options I see are:

* keep them where they are and accept the lint flag;
* ship them from another location (e.g. `files/licenses/LICENSE-OFL-1.1.txt` — reuse skips `LICENSE*`-named files — with the matching `files/CMakeLists.txt` and `Product.wxs` tweaks), which makes lint fully green;
* drop them from the repo and point THIRD-PARTY-NOTICES.md at the canonical texts — Zlib does not require shipping the text with binaries, and OFL only matters where the Roboto font is actually distributed (the web bundle, which carries its own license report).

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ and platform packaging help.
 
 NSClient++ is dual-licensed — you may use it under **either** the
 **Apache License 2.0** **or** the **GNU General Public License, version 2.0 (only)**,
-at your option (`SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-only`). See
+at your option (SPDX: `Apache-2.0 OR GPL-2.0-only`). See
 [COPYING](COPYING) for the summary, the [`LICENSES/`](LICENSES) directory for the
 full license texts, and [THIRD-PARTY-NOTICES.md](THIRD-PARTY-NOTICES.md) for
 bundled third-party components.
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -42,3 +42,42 @@ path = ["installers/ui/WixUI_MondoNSCP.wxs", "installers/installer-NSCP/WixUI_en
 precedence = "override"
 SPDX-FileCopyrightText = "Microsoft Corporation"
 SPDX-License-Identifier = "CPL-1.0"
+
+# CMake modules for C#/.NET/Mono, based on the GDCM project's (BSD-3-Clause).
+# override: the attribution line is wrapped mid-sentence, which the reuse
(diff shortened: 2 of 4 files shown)


mickem:
A few comments here.

1. wixca: This is likely 3.1 or older as wix 3.1 was the version I used until recently so your change is correct.
2. All the "dot-net" stuff is currently dormant as it is not used. So one option would be to "just delete them" it. I haven kept it as I want to enable it at some point but it is not built, not shipped and not used currently so I could easily find anything I need in the git history if need be

To your question:
They are One thing thats hard is to differentiate with source code and shipped binary on linux and windows builds. On windows we ship a lot of binaries we dont do that on Linux. So not sure whats the correct way to go here honestly.
But there is a separate license for windows build so thats likely where they should live.

mickem:
A followup the wix dll is actually version 2.0.5325.0 so it is CPL-1.0

Fantu:
Thanks for checking the actual DLL version! So wixca.dll is WiX 2.x → CPL-1.0, which is exactly what this PR annotates — no changes needed on that point.

On the dormant .NET/C# CMake modules: fine either way. The BSD-3-Clause annotation is correct for as long as the files exist, so deleting them can be a separate cleanup whenever you feel like it, and this PR stays valid in the meantime.

On OFL/Zlib: agreed, the Windows build's own license document is the right home. REUSE only covers files that are actually in the repository, so attributions for binaries that are shipped but not in-tree (zlib, the Roboto font, etc.) belong to the installer/distribution notices rather than REUSE.toml. The two "unused licenses" warnings are harmless in the meantime, so that can be a follow-up at your leisure.

PR #1350, opened 2026-07-13
Feature/more windows tweaks

mickem:


Fantu:
per-process CPU collector running it continuously impact performance? (Although I assume it's a small percentage on recent systems.)
I'd also like to add event checking to the default checks I use, but given how inefficient the Windows event system is and how much it takes and takes up when debugging, I'm not using that either. I haven't even tried the NSCP checks. Maybe I'm wrong and it might be less expensive than I imagine?

mickem:
Everything running continuously impacts performance. So on a normal machine you likely wont notice on a low-spec one with heavy load you might. But this is doing the same thing task-manager is doing so not overly so.

but this is why it is disabled by default and you have to manually enable it: See https://github.com/mickem/nscp/pull/1350/changes#diff-4479ad217e343cdb655e5d6ef41b666f9fc67e6b69e3aace43078764b5401465R445 and https://github.com/mickem/nscp/pull/1350/changes#diff-230d3780cb9afdbeeb35b28f6b49e50483724918efbde7844ad211891bd39bf4R166-R168

There is a similar pattern for the check_process_history commands as well.

As for event log: with bookmark feature (or scanning recent events) it should not be a massive performance issue unless you have some apps which logs very very heavily.

The biggest problem with eventlog is that many components and apps write status messages as errors so it can sometimes be hard to filter out noise.

PR #1371, opened 2026-08-06
Fixed #1368 disable = cpu_frequency also disables CPU load sampling, silently stalling check_cpu

### LINKED ISSUE #1368: `disable = cpu_frequency` also disables CPU load sampling, silently stalling check_cpu
**Describe the bug**

The `disable` setting of the CheckSystem collector (`[/settings/system/windows]`) is parsed with substring searches instead of exact token matching. Since `"cpu"` is a substring of `"cpu_frequency"`, setting `disable = cpu_frequency` silently disables the CPU load sampling as well.

In `modules/CheckSystem/pdh_thread.cpp` (master, around lines 330-341):

```cpp
bool disable_cpu_frequency = disable_.find("cpu_frequency") != std::string::npos;
...
bool disable_cpu = disable_.find("cpu") != std::string::npos;
```

Both evaluate to `true` for `disable = cpu_frequency`, so there is no way to disable the CPU frequency collector alone.

The consequence is not limited to the frequency data. With `disable_cpu` set, the collector skips `get_cpu_load_per_core()` / `get_cpu_load_total()` (`pdh_thread.cpp:375-380`), so `cpu.push(load)` (`pdh_thread.cpp:179`) is never reached and the rrd buffer that backs `check_cpu` stops receiving samples. `check_cpu` does not report an error in this state: it keeps answering from whatever is left in the buffer, so a monitoring system receives plausible-looking CPU values that are no longer being updated.

The same code is present in 0.14.1 (`pdh_thread.cpp:264` for `disable_cpu`, `:300` for the guard, `:104` for `cpu.push`), so the released version is affected too.

`"cpu"` / `"cpu_frequency"` is currently the only colliding pair among the documented values (`battery,cpu,handles,network,temperature,cpu_frequency,os_updates,metrics,pdh`), but the parsing would break the same way for any future value that contains an existing one.

**To Reproduce**

1. In `nsclient.ini`, set:

```ini
[/settings/system/windows]
disable = cpu_frequency
```

2. Restart the NSClient++ service.
3. Look at the log: both messages appear, although only the frequency collector was requested to be disabled.

```
WARNING: cpu frequency checking is disabled
WARNING: cpu checking is disabled
```

4. Run `check_cpu` repeatedly under varying load: the reported values no longer follow the actual CPU usage.

**Expected behavior**

`disable = cpu_frequency` disables only the CPU frequency collector and leaves CPU load sampling (and therefore `check_cpu`) untouched. Each entry in the comma-separated list should be matched as a whole token, e.g. by splitting the value on commas, trimming, and comparing exactly, rather than by searching for the token inside the raw string.

**Independently of the parsing fix, it may be worth considering whether `check_cpu` should report an error, rather than stale buffer contents, when the sampling that feeds it has been disabled.**

**System**

- OS: Windows Server
- OS Version: Windows Server 2019
- Version: 0.14.1 (x64 MSI); code inspected on master at 303df101

**Additional context**

Found while trying to reduce the cost of the 1 Hz collector loop on a virtualized RDS host, where the temperature, CPU frequency and battery collectors have nothing useful to report. The intent was to disable those three; disabling `cpu_frequency` took `check_cpu` down with it, which was only noticed because the log line for `cpu` appeared without having been asked for.

mickem:


Fantu:
thanks for the fix
looking the commit I saw a strange changes that seems not related: WIX_ROOT_DIR-NOTFOUND
I don't even see any other matches like its setting in the repository: https://github.com/search?q=repo%3Amickem%2Fnscp+WIX_ROOT_DIR-NOTFOUND&type=code

mickem:
Thanks for letting me know... I really should remove the script it has not been used I many years and always updates which is super annoying...

Fantu:
looking https://github.com/search?q=repo%3Amickem%2Fnscp+re-generate.bat&type=code it turns out `installers/common/re-generate.bat` is generated (in build) as you said so it should better remove it from the repo

anyway I not found a call of re-generate.bat, where is used?

PR #1375, opened 2026-08-06
feature/add_disk_write_check

mickem:


modules/CheckDisk/check_disk_write.cpp
```diff
@@ -0,0 +1,218 @@
...
+  // A recognizable, deterministic pattern: easy to identify if a test file is
+  // ever left behind, and verifiable on read back.
+  static const std::string pattern = "NSClient++ disk write test data. ";
+  std::vector<char> chunk(64 * 1024);
+  for (std::size_t i = 0; i < chunk.size(); ++i) chunk[i] = pattern[i % pattern.size()];
+
+  const std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
+  bool created = false;
+
+  FILE *file = fopen(path.c_str(), "wb");
```
mickem: There is a TOCTOU window between the `exists()` check here and the `fopen(path, "wb")` below: a file created in between is silently truncated and then deleted, which contradicts the documented guarantee ("never touch a file we did not create"). Opening with the C11 exclusive mode closes the race:

```cpp
FILE *file = fopen(path.c_str(), "wbx");
```

`"x"` is supported by glibc and by the Windows UCRT, and makes `fopen` fail with `EEXIST` when the file already exists — you could then drop the `exists()` pre-check entirely and map `EEXIST` to the "file already exists (refusing to overwrite it)" message. Low practical risk for a monitoring probe path, but it's a one-line fix that makes the guarantee real.

---

modules/CheckDisk/check_disk_write.cpp
```diff
@@ -0,0 +1,218 @@
...
+    file = fopen(path.c_str(), "rb");
+    if (file == nullptr) {
+      issues.push_back("failed to read file back: " + last_error());
+    } else {
+      // Mirror the write loop chunk for chunk so the verify buffer lines up.
+      std::vector<char> read_buffer(chunk.size());
+      for (long long remaining = size; remaining > 0 && issues.empty();) {
+        const std::size_t count = remaining < static_cast<long long>(chunk.size()) ? static_cast<std::size_t>(remaining) : chunk.size();
+        if (fread(read_buffer.data(), 1, count, file) != count) {
+          issues.push_back("failed to read file back: file is shorter than what was written");
```
mickem: A short `fread` is not necessarily a short file — a mid-read I/O error also returns a short count. Since this check exists precisely to catch flaky disks, it's worth distinguishing the two with `ferror()`:

```cpp
if (fread(read_buffer.data(), 1, count, file) != count) {
  issues.push_back(ferror(file) ? "failed to read file back: " + last_error()
                                : "failed to read file back: file is shorter than what was written");
  break;
}
```

Same applies to the trailing one-byte read below (a read error there currently reports "file is larger than what was written").

---

modules/CheckDisk/check_disk_write.cpp
mickem: Minor, Windows: `fopen`/`std::remove` with a narrow path go through the ANSI codepage, so a non-ASCII path (e.g. `D:\Übung\probe.dat`) will fail or hit the wrong name. Elsewhere in this module Win32 calls convert with `utf8::cvt<std::wstring>` (e.g. `check_uncpath.cpp`, `filter_win.cpp`) — though the checksum code in `filter.cpp` has the same narrow-`fopen` limitation, so this is a pre-existing pattern rather than a regression. Fine to leave as-is; noting it in case you want `_wfopen`/`_wremove` behind the existing `#ifdef WIN32`.

---

modules/CheckDisk/check_disk_write_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // Thresholding on total_time also emits it as perf data.
+  ASSERT_GE(response.lines_size(), 1);
+  bool has_total_time_perf = false;
+  for (const auto &perf : response.lines(0).perf()) {
+    if (perf.alias().find("total_time") != std::string::npos) has_total_time_perf = true;
+  }
+  EXPECT_TRUE(has_total_time_perf);
+}
+
+TEST(CheckDiskWritePerform, ReportsMismatchWhenTheFileChangesUnderneath) {
```
mickem: The test name promises a read-back-mismatch scenario but the body (as the comment admits) verifies the happy path. A name like `CheckDiskWritePerform.HappyPathPopulatesResultFields` would stop a future reader from believing the mismatch branch is covered. (The mismatch branch *is* effectively exercised at the rendering level by `StubbedFailureRendersTheIssues`, so renaming rather than expanding seems fine.)

---

docs/docs/reference/check/CheckSystem.md
mickem, line 1: This CheckSystem docs addition (the `check_cpu` / `disable` interaction note) looks unrelated to `check_disk_write` — possibly a leftover from another branch. Intentional to bundle it here, or should it ride in its own PR?

---

mickem (commented):
Reviewed the new `check_disk_write` command. Overall this is a solid, convention-clean addition — I'd merge it once the small items below are considered.

**What's good**
- Follows the project conventions end to end: `module.json` command entry, generated-glue-compatible method on `CheckDisk`, win/unix CMake wiring, `_desc.md` + `_samples.md` docs, gtest unit tests **and** integration tests on both the REST path (`checkdisk-commands.test.ts`) and the client-query path (`checkdisk-unix.test.ts`). No boolean options, so the `bool_switch` REST trap doesn't apply.
- The injectable `write_tester` split cleanly separates filesystem behaviour from option parsing/rendering, and the tests use it well (stubbed failure + stubbed timing/perf-data cases).
- The `${status}: ${list}` top-syntax correctly triggers the OK-renderer blanking in `cli_helper::parse_options_post` (it matches on `{list}` as well as `(list)`), so the per-item rendering works for every status — verified against `cli_helper.hpp`.
- Error accumulation is careful: `errno` is captured immediately after each failing call, the file is deleted even when the write/read failed, and thresholds in the integration tests are pinned so CI disk speed can't flake them.

**Items raised inline**
1. **TOCTOU on the exists-then-create guarantee** — `fopen(path, "wbx")` would make "never touch a file we did not create" actually hold (main suggestion).
2. Short `fread` reported as "file is shorter than written" even when it's a read I/O error — `ferror()` distinguishes them.
3. Narrow `fopen`/`remove` on Windows won't handle non-ASCII paths (pre-existing module pattern, non-blocking).
4. Misleading unit-test name (`ReportsMismatchWhenTheFileChangesUnderneath` tests the happy path).
5. The `CheckSystem.md` `check_cpu`/`disable` docs hunk looks unrelated to this PR — leftover from another branch?

Also worth noting (no change requested): `size` has no upper bound, so e.g. `size=100G` will block a check thread for however long the write takes — arguably that's the operator's own choice for a probe like this, and disk-full is handled gracefully, so I don't think it needs a cap.

---

PR #1377, opened 2026-08-07
Feature/added sql server checks t2

mickem:


Fantu:
@mickem Thanks for this other helpful addition.

PR #1397, opened 2026-08-12
Add CheckMySQL module for MySQL, MariaDB and Percona servers

mickem:
New cross-platform check module built on MariaDB Connector/C (LGPL), which speaks the native protocol to the whole MySQL family:

- check_mysql: connectivity/health with version, flavor (mysql/mariadb/ percona), uptime (time-unit thresholds) and connection-pool keywords. Connection failures are UNKNOWN with the driver's error message.
- check_mysql_query: arbitrary SQL with result columns registered as filter keywords (CheckWMI-style), per-row matching and perf data via perf-syntax.

The module is optional: it builds when the connector is found (FindMariaDB, libmariadb-dev or -DMARIADB_ROOT_DIR) and is skipped with a clear reason otherwise. Only mysql_session.cpp touches the connector; the checks consume an injectable session factory, so the unit tests run without libmariadb.

Notable behaviour: TCP is forced unless socket= is given (the connector silently turns host=localhost into a socket connection), and plugin-dir= supports MySQL 8's caching_sha2_password auth plugin from non-default locations. Verified live against MariaDB 11 and MySQL 8.4; the integration suite starts a MariaDB testcontainer and skips itself without docker or without the module built. MSI entries are present but commented until the connector is part of the Windows build dependencies.

Assisted-by: Claude Code:claude-fable-5

### DIFF
--- a/.github/actions/mariadb/action.yaml
+++ b/.github/actions/mariadb/action.yaml
@@ -59,10 +59,20 @@ runs:
     # extra file. auth_gssapi_client is dropped: it is the only plugin that
     # would drag in an external dependency, and CheckMySQL does not offer
     # Kerberos auth.
+    #
+    # CMAKE_SYSTEM_VERSION=10.0 selects the newest installed Windows 10 SDK.
+    # Without it CMake reports "Selecting Windows SDK version 8.1 to target
+    # Windows 6.2.9200" (the runner has the Windows 8.1 SDK installed for the
+    # legacy XP/cryptopp build path, and the runner reports itself as 6.2) and
+    # the connector's Schannel backend then fails to compile: schannel.c needs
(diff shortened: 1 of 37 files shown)


Fantu:
Thanks for this other useful addition.
I'll need it on Linux, but unfortunately I haven't had time to continue work on nscp on linux, and I don't think I can even do it these days.
I'm currently trying to add and improve the Windows device monitoring, which is the most problematic. Lately, several issues have been costing me a lot of time debugging, so I'm trying to add checks that will help detect some difficult cases.
When I'll have a bit of peace, I'll get back to focusing on the Linux monitoring. Even though the problems are rarer and often easier/quicker to resolve on linux, monitoring is still useful. I'm already using icinga2 monitoring on Linux, but with other tools. The ultimate goal in the future would be to use nscp with icinga push passive checks on all Windows and Linux systems.

PR #1412, opened 2026-08-16
Windows security and system checks: activation, file ACLs, W32Time, print jobs and firewall rules

mickem:
Five new Windows checks and one gap closed in an existing one, each with unit tests, integration tests and documentation.

## New checks

**`check_activation`** (CheckSecurity) — Windows activation and licensing state: license status, remaining grace or KMS renewal period, channel and genuineness, from `SoftwareLicensingProduct` plus `SLIsGenuineLocal` in `slc.dll` (loaded dynamically, no new link dependency). Critical when not licensed — which covers unlicensed, in-grace and the "activate Windows" notification state — and warning while a countdown has less than 30 days left. A permanently activated host reports `grace_days = 0`, so the warning only fires where a period is actually running.

**`check_file_security`** (CheckSecurity) — owner and DACL of files, folders and service binaries, alerting on world-writable paths and unexpected owners. `service=<name>` resolves the image path through the SCM, so the check follows the binary a service actually runs.

One deviation from the obvious implementation: `GetExplicitEntriesFromAcl` returns only the ACEs set directly on an object, and a child of a world-writable folder came back with an *empty* ACL. The ACL is walked with `GetAce` instead, so inherited access — the usual way these problems arrive — is judged like any other, and marked `~` in the rendered DACL.

**`check_w32time`** (CheckSystem) — what the Windows Time service itself thinks: whether the machine follows a time source at all, which one, the computed offset and round-trip delay, the configured peers, and how long ago the clock was last recorded as good. This is the inside-out counterpart to CheckNet's `check_ntp_offset`; a domain member whose time hierarchy has broken keeps answering with a plausible clock for hours while Kerberos validation is already failing.

`synchronized` ranks its evidence rather than guessing: a stopped service or `Type=NoSync` settles it; otherwise the service's own answer decides when it can be had (`W32TimeQuerySource` needs privilege, which the agent has as a service); when it cannot, `NTP Client Time Source Count = 0` decides. With neither, the check reports configured intent instead of alerting forever.

**`check_printjobs`** (CheckSystem) — one row per queued job: printer, document, owner, size, pages printed of total, priority, submit time, age, and the spooler `StatusMask` decoded into both words and individual keywords. Critical on the states the spooler cannot clear itself (error, blocked device queue, waiting for a person); a paused job is deliberately not critical.

**`check_firewall_rules`** (CheckSecurity) — one row per firewall rule from `INetFwPolicy2::Rules`. `expect=<name>` asserts a rule is in effect, failing whether it was deleted or merely switched off and saying which. `any_any` marks an enabled inbound allow rule that restricts neither address nor port — a keyword rather than a default threshold, because a normal client legitimately has ~120 of them (one per packaged app).

## Extended

**`check_printqueue`** gained the device inventory behind each queue — `driver`, `port`, `location`, `share`, `server` and the default/shared/network flags — usable as filters (`filter=shared = 1`). When a queue looks healthy and printing still fails, the answer is usually that the driver or port changed underneath it.

## Bugs found and fixed along the way

- **`oldest_job_age > 30m` never worked.** The documentation has always promised it, but the keyword was a plain int, so `30m` parsed as the number **30** and the threshold fired on any queue holding a job for half a minute. Duration keywords need a converter against a custom int type (the mechanism `check_uptime` uses); it now lives in `duration_keyword.hpp` and is shared. Bare numbers still mean seconds, so existing configuration is unaffected.
- **`check_w32time` reported a false OK** on a machine following nothing at all: the source was unknown (access denied unprivileged) and the check presented the *configured* peer as if it were the live one. Caught by starting the service on a test machine; fixed by the evidence ranking described above.

## Notes

- All five checks are Windows-only and report the standard "not supported on this platform" contract elsewhere; the platform-neutral halves were compiled with gcc as well.
- Optional numeric values (`offset`, `delay`, `frequency_adjustment`, …) are real optionals rather than `-1` sentinels — the clock frequency adjustment is legitimately negative. They render as `unknown`, compare false against every number so no threshold can fire on a missing measurement, and emit no perfdata.
- Committed reference documentation (`docs/reference/*.yaml` and the rendered Markdown) is regenerated in separate `docs(...)` commits.

## Testing

- **CheckSecurity**: 39 unit tests in `check_posture_test` plus 8 in `check_firewall_test`; 29 integration cases pass. (Two pre-existing `check_group_members` cases fail on the test machine because they hard-code the English group name `Administrators` — unrelated to this branch.)
- **CheckSystem**: 425 unit tests and all 68 integration tests pass.
- Verified against live state where possible: real firewall rule set (631 rules, cross-checked with `Get-NetFirewallRule`), a real queued print job, real printer inventory, the running Windows Time service, and a genuinely world-writable directory created for the test.


mickem:
Addressed the code-review findings in four follow-up commits (c9e8469b..0e2dd09b):

**Fixed (7 of the 8 correctness findings):**
- `check_file_security`: generic↔specific rights now cancel via a `MapGenericMask`-equivalent before deny subtraction; a failed `GetAclInformation` reports the unreadable-descriptor critical instead of a clean pass; UNC service image paths are no longer corrupted by drive-letter prefixing.
- `check_firewall_rules`: a user `filter=` can no longer swallow the hole row for a missing `expect=` rule (filters are wrapped as `(<filter>) or present = 0` when expectations are in play), and a duplicated `expect=` name no longer fabricates a bogus "exists but is disabled" hole.
- Duration keywords: the `-1` sentinel is comparable again (plain signed integers bypass the time parser, the same guard the CheckMSSQL/MySQL/Docker copies carry), and `last_sync_age` takes duration literals like its siblings.
- WMI: `Win32_Printer` and `Win32_PrintJob` queries name their columns again instead of `SELECT *`.

**Cleanups:** the converter guard is in the one shared header and `check_uptime`'s in-module copy now uses it; consolidating the remaining per-module copies (CheckSystemUnix, CheckDisk, CheckMSSQL, CheckMySQL, CheckDocker) is deliberately left as follow-up rather than widened into this branch.

**Disputed (1):** the claim that `synchronized` gives a false OK on localized Windows because "Local CMOS Clock" is translated. Direct evidence against: on this Swedish-locale machine — where `w32tm` errors come out in Swedish — the live source string is reported as `Local CMOS Clock` verbatim; it is a provider-internal string, not a MUI resource. Documented in a code comment rather than changed, since the plausible mitigation (letting `time_sources = 0` override a live source) would false-alarm Hyper-V guests synchronized through the VMIC provider.

All verified: 43 + 428 unit tests, both integration suites green, and each fix exercised live (the filter+expect hole now survives — 417 rows = 416 enabled + 1 hole; `Default` is accepted as a named WQL column; `oldest_job_age = -1` fires on empty queues).

Issue #1410, opened 2026-08-14
Add an upgrade-notes document in the repository collecting breaking changes across versions


Fantu:
**Is your feature request related to a problem? Please describe.**

NSClient++ has no place in the repository that answers the single question every
upgrade starts with: **what changes between the version I am running and the version
I am installing require me to do something, or silently change how my existing
configuration behaves?**

The recent development pace makes this much more pressing: a lot of new checks,
fixes and behaviour changes across several releases in a short time (0.13.0 → 0.16.1
in roughly two months), so there are more versions to cross and more changes that can
bite. Those releases do contain genuinely good upgrade information — the "⚠️ Upgrade
notes" section at the bottom of the release notes — but it lives *only* in the GitHub
release body of each tag, so the number of places to consult grows with every release
while the answer stays scattered.

That said, the need does not come from the pace, it is only amplified by it. I have
been using NSClient++ for many years and the same difficulty was there when releases
were rarer and breaking changes fewer — arguably it is even easier to get caught
then, because a slow-moving project is exactly the one you install and forget. That
is the normal life of a monitoring agent: it is set up once, it works, nobody touches
it for months or years, and then someone upgrades it across a large version gap. At
that point it makes little difference whether the gap contains twenty releases or
two: what matters is whether the handful of changes that require intervention is
written down somewhere you can read in one go. So this would be worth having even if
development slowed down tomorrow — documenting breaking changes and significant
behaviour changes in a durable, versioned place seems to me a basic expectation for
any serious project, independently of how fast it moves.

Today that means:

- opening the GitHub releases page and reading every single release body one by one —
  every one of them, because the version number tells you nothing about what is safe
  to skip: a minor or patch-level bump is not a guarantee of "no breaking changes",
  and in practice substantial changes do show up there too. Some releases turn out to
  have no upgrade notes at all, but you only find that out by opening them. There is
  no shortcut: you read them all, or you risk missing exactly the one that matters;
- mentally merging the notes of N releases into one list, and hoping nothing was
  missed;
- doing it online — there is nothing in the source tarball, in the installed
  package, or on the documentation site that answers "what breaks if I jump from
  0.13.0 to 0.16.1?".

And getting it wrong has consequences of very different severity: the agent may not
work at all after the upgrade, or work only in part, or work while quietly producing
different results than before. In every case you find out afterwards, by debugging,
instead of before, by reading. I have come across changes of this kind over the years,
and I have had a hard time precisely because a document like the one I am asking for
[...]

PR #1419, opened 2026-08-19
docs: restructure the upgrading page by impact and fix attribution

Fantu:
Follow-up to #1410.

## Why

Every item of a release sits in one flat bullet list, so a mandatory action, a behaviour change nobody has to act on and a purely opt-in addition all look alike. That is hardest on the two readers the page exists for: an operator jumping several versions, who wants only the things that will not fix themselves, and anyone handing the page to an assistant, which cannot tell which items apply to a given configuration without re-deriving all of it from the source.

## What changed

Each release is split into **Requires action**, **Changes behaviour (no action)** and **New and opt-in (no action)**; empty sections are omitted, so a release that asks nothing of you is visibly short. Each item starts with the module, setting or command it concerns, so searching the page for a name you have configured finds everything that touches it.

I established the introducing release for every item from the commits (`git log -S…`, then `git tag --contains`) rather than from the release notes, which surfaced a set of attribution problems:

| Problem | Fix |
|---|---|
| 0.11.32, 0.12.0, 0.12.1, 0.12.2, 0.13.0, 0.14.0, 0.15.0, 0.16.0 have upgrade-relevant changes but no section | added, with the items placed by commit |
| Everything under 0.12.3 belongs to 0.12.0-0.12.2 (0.12.3 has 19 commits and no upgrade-relevant change of its own) | 0.12.3 removed, items moved to the release that introduced them |
| `[/paths]` → `boot.ini` and `disable admin user` listed under both 0.12.5 and 0.12.6 | kept in 0.12.5 (`fbdfe257`, `cccc14e4`); 0.12.6 only renamed the `--path-override` CLI flag (`a37bb258`) |
| The web-server-refuses-cleartext / `install-ui` / FHS-layout items are 0.13.0 (`82094007`, `c1ae115d`, `2a945286`), re-announced in 0.14.0 | listed once, under 0.13.0 |
| Setting name given as `cache allowed host` | it is `cache allowed hosts`, plural (`socket_settings_helper.hpp:98`) |
| "Replace clients that call `/auth/token` or `/auth/logout`" | both endpoints still exist (`legacy_controller.cpp:28-30`), behind the WEB `legacy` permission. The 0.12.2 change (`340b8db1`) is that passwords and tokens are no longer accepted in the query string |
| Items that never made it onto the page | recovered, e.g. the `check_nt` literal `None` password no longer authenticating (`6fcb6e44`), `cache allowed hosts` becoming a real boolean (`b3337534`), Nagios range syntax in perfdata (`86da6b5b`), and the mixed `warn=`/`crit=` evaluation changes (`24ffb992`, `dd8024ae`) |

Pre-releases get their own sections rather than being folded into the next stable one: people do install them, and folding is what put 0.12.0-0.12.2 under 0.12.3 in the first place.

`CLAUDE.md` documents the section split, and asks for the introducing release to be established from the commit when in doubt.

## One open question

**`insecure = true` on NRPEServer logging at ERROR** is in the 0.12.6 release notes, but I could not place it: no commit in `0.12.5..0.12.6` touches the log level, and the message in `include/net/socket/connection.hpp:306` has been `log_error` since 0.4.3.147. I left it off rather than file it under a release I could not confirm — happy to add it wherever it belongs.

### DIFF
--- a/docs/docs/setup/upgrading.md
+++ b/docs/docs/setup/upgrading.md
@@ -1,44 +1,41 @@
 # Upgrading
 
-What to do when upgrading NSClient++, newest release first. Most upgrades are
-in place — defaults are preserved and the default install is usually unaffected
-— but the items below change observable behaviour or want a configuration
-touch. Read the entries between the version you are on and the version you are
-moving to.
+What to do when upgrading NSClient++, newest release first. Read the entries
+between the version you are on and the version you are moving to. Most upgrades
+are in place — defaults are preserved and the default install is usually
+unaffected.
 
-Each entry carries an icon for the area it touches. 🔒 is the one that
-matters: those entries are security-relevant, and the
-[Security notices](../security/notices.md) page tracks them in one place. Full
-per-release detail lives in each
+Every release is split by how much it asks of you, so you can stop reading at
+the point where a section stops applying:
+
+| Section | What it means |
+|---|---|
+| **Requires action** | An existing install needs a change from you — a setting to add, a threshold, client or alert to review. Every item names its condition up front, so skip the ones that do not apply. |
+| **Changes behaviour (no action)** | Something observable changed, but the upgrade is in place. Read it if output, perfdata or alerting looks different afterwards. |
+| **New and opt-in (no action)** | Additions that stay off until you turn them on. Nothing breaks if you skip the section entirely. |
+
+Each item appears **once**, under the release that actually introduced it —
+including the releases marked as pre-releases, since those are installed too. A
+section is omitted when a release has nothing under it, and a release with no
+upgrade-relevant change is not listed at all. **Skipping several versions?**
+Read every *Requires action* section between your version and your target
+first — together they are the complete list of things that will not sort
+themselves out. Each item starts with the module, setting or command it
+concerns, so searching this page for the name of something you have configured
+finds everything that touches it.
+
+Each entry carries an icon for the area it touches; 🔒 is the one that matters —
+those entries are security-relevant, and where a notice exists the item links to
+it on the [Security notices](../security/notices.md) page, which tracks them in
+one place. Full per-release detail lives in each
 [GitHub release](https://github.com/mickem/nscp/releases).
 
 ---
 
 ## 0.18.1
 
-- 🔒 **Icinga API submissions now honour the configured `timeout`, and
-  credentials no longer reach the trace log.** The `IcingaClient` module's
-  HTTP calls previously waited forever — the target's `timeout` setting
-  (default 30 s) was read but never applied — so a stalled Icinga endpoint
-  could silently wedge passive-result submission; set `timeout = 0` on the
-  target if you depend on the old unbounded wait. The same pass masked
-  `password`/`token` values in the trace-level target dump (this also covers
-  the other client modules sharing that machinery) and added a log message
-  when an `https` submission runs with certificate verification disabled
-  (`verify mode` empty or `none`). See
-  [Security notices](../security/notices.md#security-hardening-across-the-clients-scripts-and-filter-framework).
-- 🔒 **Filter expressions are now bounded in length and nesting depth.** A
-  `filter` / `warning` / `critical` expression — and a `%(...)` expression
-  placeholder inside a syntax template — longer than **1024 characters** or
-  nested more than **64** parentheses deep is now rejected at parse time
-  instead of being evaluated, failing with a clear "exceeds the maximum
-  length/depth" error. The where-parser and the expression evaluator both
-  recurse with the shape of the input, so an unbounded or deeply nested
-  expression could exhaust the stack and crash the agent. Real filters are a
-  small fraction of these limits, so the default install and every normal
-  configuration are unaffected — only a pathologically large or deeply nested
-  expression is refused. See the
-  [security notice](../security/notices.md#security-hardening-across-the-clients-scripts-and-filter-framework).
+### Requires action
+
 - 🔒 **CheckExternalScripts hardening.** A hardening pass over the external
   scripts module tightened several rough edges: the `ext-scr install` argument
   lockdown now writes to the setting the module actually reads (previously it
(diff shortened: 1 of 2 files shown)


docs/docs/setup/upgrading.md
```diff
@@ -1,128 +1,373 @@
...
+- **`check_disk_health` and `check_disk_io` perfdata series are renamed.**
+  `free_pct`, `queue_length` and `percent_disk_time` carry their own label
+  suffixes instead of sharing the bare drive label, so Graphite and InfluxDB see
+  new series — the old collapsed series held the wrong metric anyway. Icinga
+  shows two correctly named entries where it showed two under one name.
+- **Perfdata is omitted for values that were not measured** rather than plotted
+  as `-1`, so those series appear and disappear instead of carrying a fake
+  floor.
+- **CheckMySQL** is built only where MariaDB Connector/C is available, and its
+  MSI entries ship commented out.
```
mickem: Not sure which version but CheckMySQL is not commented out and shipped in the installer...

mickem:
Thanks for this...

I am working remotely so just looking at this briefly...

Regarding your question this is the commit which is referd: https://github.com/mickem/nscp/commit/03809914#diff-4680b335407df86653bf4e430227472f160f42327fab110d3cc5bfb50c7e831fR114-R123
But I havent looked at which release it is in.

A though... A few comments say things along the lines of "...Module only on windows..." which likely was true "at the time", but many modules gets parity "at a later point" so might not be true "now".
Is it significant that modules are linux or windows only?

Fantu:
I think is useful to know if the entry is related to a single platform.
In many cases, it can be helpful to review many things in a short amount of time. Not having that documentation, or if it's too long or complex, discarding it can lead to more unexpected issues during the upgrade or afterward.
If you could have good, accurate, and clear documentation, where you can quickly select only the things relevant to your current installation, that would be ideal.
Having additional, well-written notes and additions is also a plus.
Then, to add ir changes checks, often need to consult the detailed documentation for the updated version (or the documentation for the version you're using). So for additions/new things (as for some other things) you can also avoid being too long there.

I don't know if I'm making myself clear and if we'll be able to achieve a good result.

mickem:
Yes I agree with that it is valuable information. 

I just meant that it is information that will change over time and it will change on "check-level" more than module level. 
So my question is more if I am upgrading from 1.2.3 to 4.5.6 would I care that a certain module was windows only in 2.2.2 and then got linux support in 3.3.3? I might want to know that it exists but which checks it has and what is supported might change over time.

The docs are currently built "platform based" so it would be possible to for instance add a column in the https://nsclient.org/docs/reference/ to include which platforms the various modules/checks are avalible on.

So would solve the same problem?
The benefit of that is that it keeps upgrade focused on whats changed and it maintains it self as the docs are built based on the code.

Fantu:
The situation is actually a bit complex. Checking the individual entries to see if they're platform-specific can help, but when updating after a long time, you might find yourself reading about several things (perhaps many) that have changed over time in subsequent versions (also not only related to platform). In most cases, I assume you'd look at the current documentation (especially the additions), but for those who want to rely heavily on it, it's not ideal.

I'm not sure what could be done to create something optimal for both main situations; perhaps it won't work, and you'll have to choose the option that might be best.

mickem:
As it costs nothing I added os support to the index page here: https://nsclient.org/docs/reference/ (and yes I noticed I accidentally added a forthcoming module, I did not see it end up in the index, but it will be there "soon" :)

mickem:
Lets review this now that we have a base line for the "proper update entries".

Check the last two versions on the upgrade page and let me know how we should change them: https://nsclient.org/docs/setup/upgrading/

Fantu:
Sorry, I've been busy with other things the past few days.
This is a long and complex issue. I had Claude update it based on the previous guidelines, but I haven't had enough time to look it over. Below is the response Claude generated for the PR:

--------

Rebased on main, so the page now carries 0.17.0, 0.18.0 and 0.18.1 and your area icons.

I checked the last two releases first: **attribution is right** — every item in 0.18.0 and 0.18.1 sits under the release that introduced it (verified with `git tag --contains`, not from the release notes), so nothing had to move. What I changed on them:

- **Split by impact**, like the rest of the page. 0.18.1 comes out as 9 *Requires action* / 9 *Changes behaviour*; 0.18.0 as 7 / 11 / 2; 0.17.0 as 9 / 6 / 1. The single-item 0.16.3 / 0.16.4 sections got a heading too, so the whole page is filed one way.
- **Three entries covered more than one change** and could not be filed as they stood, so each is now one item per change: NSCA hardening (empty encryption password = fix it; perf-data and wire-field validation = nothing to do), NRPE (metachar guard vs. the new opt-in `expose version`), WEB (a name starting with `-` must be renamed; the rest needs nothing).
- **Two changes had a security notice but no upgrade entry**: the SMTP reply scrubbing/buffer caps, and Graphite now logging a failed metrics submission instead of dropping it. Added.
- Dropped the `---` rules between releases — one of them sat in the middle of 0.18.0 and read as a release boundary.

I left the prose alone. Which leaves the one thing I would still change, and it is your call:

**Length.** 0.18.1 is ~180 lines on main, where 0.16.1 is 21 and 0.14.1 is 13. The page is read by someone jumping several versions, and much of the length is the archaeology of *why* the bug existed — which the release notes and the linked security notice already carry. A rule that would keep it useful: **what an operator must do, plus one sentence of why; the history stays in the release notes.** For example:

> ⏱️ **Graphite `timeout` is enforced now, as one budget for the whole submission.** The configured value never reached the connection, and the connection had no deadline anyway, so a stalled carbon endpoint could hold the submitting thread indefinitely. A target that previously completed by quietly taking longer will now fail at the configured value — raise `timeout` for a slow relay. See [Security notices](../security/notices.md).

5 lines instead of 12, with nothing an operator acts on removed. Want me to apply that pass across 0.17.0–0.18.1 in this PR, or keep the entries as written?

One smaller question: the icons have no legend, so beyond 🔒 a reader cannot tell 📤 from 📨 from ✉️. Add a one-line legend to the intro, or keep icons but only guarantee the meaning of 🔒?

Fantu:
Even with this change, I still find it difficult to find the information is needed during an upgrade (though it should help a little). If you do it manually after a long period of time, it's a nightmare. I'm assuming someone will try instead to see if there are any errors or checks that stop working after the upgrade.
I've recently been using Claude to check which changes affect the parts I use, and I only look at those (when I update the configuration to start using a new version).
Furthermore, given the constant changes and additions I'm making to my configuration, I've also added a file with the entire configuration that's the same for all the devices (that I include in the main file). Below that, I add in the main file a few generic configurations and commands whose values ​​I frequently change on individual devices, as well as any copies of commands I change exceptionally.
I also keep a comment at the top of each command with the version to more easily track if it's changed.

mickem:
Yeah, the length is what struck me the most, I had not even thought about this side of the coin as I said initially "in my mind" this was not an issue, but boy was I wrong :)
Update notes was almost as long as the release notes... 
So thinking freely here: An option would be to add some kind of filter mechanism where you can say "only security" or "only modules X, Y, Z" perhaps?
The other option is to make them shorter but then we might end up back where we started...

I'll ponder this a bit...

I also want to say that a lot of the "changes we have" are driven by two things:
1. I add a lot of test to "old code". This surfaces bugs that may not have been visible before
2. Updated security

Bot of these are (hopefully) thing that will soon go down as we have fixed "all the old bugs" and already the security issues founds are mostly exotic and theoretical at best. 
So maybe this is a "problem now" that will not exist 6 months from now?

The other thing this surfaced was this was PURE merge hell, I spend litterally hours and hours and hour rebasing and merging these changes.
So having a single markdown file in the source is not feasable. So I think a better option would be to source this as `/docs/upgrade-notes/<version>/name.md` and build the upgrade.md from that.

Anyways, I see the problem, and the pain and I am actively thinking of how to bridge the gap.
But I will explore if I can  make some way to query the information instead of formatting the markdown so it is suitable to everyone.


And massive thanks for taking the time!

mickem:
What do you think of this:

image

Fantu:
The restrictive filter seems interesting, although restricting by modules might not be simple and might include things you don't use (but for checks you do use instead, I think it risks being too long a list to compile and becomes counterproductive).
In addition to filtering security only, I suspect it would be more useful to filter things that might require user modifications.

mickem:
How about this (new option with requires human intervention): 
image

Fantu:
I think that filter will be useful, probably the name "may needs action only", add "may" to try not to scare people if many results come up, not all of them could concern the user and even if they do concern the things used, changes are not necessary in all cases

mickem:
New concept is live here: https://nsclient.org/docs/setup/upgrading/ and https://nsclient.org/docs/security/notices/

Anything we want to tweak and//or change/enhance?

Fantu:
I tried it a little while ago to see notes for updating my configuration and internal documentation for moving from 0.17.0 to 0.19.0.
By selecting "may need action only," I got to 27 of 93 entries,
and by deselecting the parts I don't use, I got down to 10 of 93 entries.
It basically seems better, although I haven't checked it thoroughly. I don't have time for an in-depth analysis, and having only an LLM do it isn't a good idea (as already seen with this PR, I had to address and correct it extensively).

Fantu:
Closing this: the per-file notes and the filter replace what this PR was trying to do, and the page can no longer be edited as a single file anyway.

I did not go through the migrated notes in depth, only a quick pass with Claude, but two things stood out that could mislead a reader today:

- `0.12.3/030-replace-auth-token-clients.md` says to replace clients that call `/auth/token` or `/auth/logout`, but both endpoints still exist (`legacy_controller.cpp`). What changed is that passwords and tokens are no longer accepted in the query string.
- Nothing mentions that `check_process delta=true` needs `process cpu = true` under `[/settings/system/windows]` since 0.15.0 (`8306eb48`); without it the check returns UNKNOWN.

The older versions have more gaps, but I don't think they are worth reconstructing (can take long time or risk errors).

PR #1421, opened 2026-08-20
fix(check_nt): restore real check_nt compatibility + integration suite against nagios-plugins

mickem:
## Summary

Integration-testing the legacy NSClient protocol against the **real** nagios-plugins `check_nt` immediately surfaced a regression: the buffer-cap hardening that shipped in 0.12.2 (3823a740) made the request parser wait for a newline terminator — but the real `check_nt` sends `<password>&<cmd>&<args>` with no terminator and waits for the reply, so every request from it has been hanging until the client's socket timeout (`No data was received from host!`) in every release since.

### Fix

`include/net/check_nt/server/parser.hpp`: end-of-read is once again end-of-request, while keeping both halves of the hardening — the 4 KiB request cap and the newline path (which consumes the terminator and leaves pipelined bytes intact) for line-oriented clients. New `CheckNtParser` unit tests in `net_test` pin the no-terminator format, the newline path, the empty chunk and the oversized-line cap.

### New integration suite

`tests/check_nt-client.test.ts` + `tests/Dockerfiles/check_nt.Dockerfile` compile `check_nt` from the official nagios-plugins 2.5 release tarball in an Alpine container (only `check_nt` and the libraries it links) and drive it against `nscp test` on port 12489, following the `nrpe-tls.test.ts` pattern:

- **Communication** (default `allow = any`): `CLIENTVERSION`, `UPTIME`, `CPULOAD` (polling out the 1 Hz collector warm-up) everywhere; `MEMUSE`, `USEDDISKSPACE`, `PROCSTATE` on Windows — asserting on check_nt's actual rendered output and exit codes.
- **Password enforcement**: a wrong password gets the generic `ERROR: Bad request.` (exit 3) with no valid/invalid oracle.
- **`allow` gating** (`allow = 'info, uptime'`): the group and the individual command answer; `MEMUSE`, `COUNTER` and `PROCSTATE` are refused before dispatch with `ERROR: Command not allowed.` (exit 3).
- **Fail-closed**: an allow list containing only an unknown token refuses even `CLIENTVERSION`.

MEMUSE and PROCSTATE are Windows-only in the suite because their server-side mappings are Windows-shaped (`check_memory type=committed` and the `legacy_state` keyword don't exist in the Unix modules) — pre-existing interop gaps, noted in the test comments.

## Testing

- `net_test`: 517 passed (incl. the 4 new `CheckNtParser` tests).
- `check_nt-client.test.ts`: 13/13 passed on Windows against a RelWithDebInfo build (Rancher Desktop docker). Before the parser fix, all communication tests failed exactly the way real deployments do.


mickem:
Added on top of the original fix + integration suite:

- **refactor(check_nt):** extracted the dispatch-free protocol logic (allow parsing, request-code → modern-query mapping, response → legacy-payload formatting) out of `NSClientServer.cpp` into `check_nt_commands.{hpp,cpp}` and unit-tested it (26 tests: group/command allow semantics incl. fail-closed, per-code argument mapping, and the easy-to-swap legacy value orderings like MEMUSE `total&used` vs USEDDISKSPACE `free&total`). No behaviour change — the integration suite passes unchanged (13/13).
- **test(check_nt):** `protocol_test.cpp` for the server `read_protocol` state machine, mirroring the existing NRPE/NSCA ones (7 tests, incl. the terminator-less request contract and exception → error-response).
- **docs(check_nt):** new `docs/samples/NSClientServer_desc.md` merged into the reference page, documenting how to secure the server (mandatory password + cleartext caveat, allowed hosts, `use ssl` vs the real check_nt) and the `allow` setting with the group table, fail-closed behaviour and a locked-down example.

PR #1427, opened 2026-08-21
Coverage reporting, and the six defects it found

mickem:
Adds gcov coverage reporting, then uses it: the low-coverage areas it pointed at turned out to be hiding six real defects, one of them a decade old.

## Coverage tooling

`-DNSCP_COVERAGE=ON` instruments every target, and `tools/coverage/run.sh` drives the whole cycle — unit suite, integration suite, merged HTML + Cobertura. Because the jest harness spawns the built `nscp` and lets it dlopen the built modules, one instrumented tree covers both layers.

Three fixes to the runner itself, all found by using it:

- **Stale tracefiles** from a previous run were merged into the current one, and gcovr aborts on an md5 mismatch as soon as a covered source has changed — after the build and both suites have already run.
- **Concurrent gcovr runs** race over `.gcov` files. gcov names its output after the source file and writes it into `--root`, and gcovr's per-directory guard is an in-process threading lock, so a second gcovr knows nothing about the first. The loud failure is a `FileNotFoundError`; the quiet one is worse, because `--gcov-ignore-parse-errors` turns a lost file into missing coverage in a report that still looks complete. That is what produced a "unit" report showing `settings_handler_impl.cpp` at 25% with its tests passing.
- Leftover `.gcov` in the repo root after a crashed run (about a hundred of them).

## Defects found

**Syslog submission has been dead since 0.4.3 (2015).** `connection_data`'s first argument is the target's settings and its second is the sender; `SyslogClient::submit` passed them the other way round, so the configured address, port, facility, severity and templates were all ignored. Against a target configured exactly as documented, the agent logs `Undefined facility:` and sends nothing. `CheckMKClient::query` had the same swap. Verified end to end against a local syslog listener: nothing before, `<6>… nscp …` after.

**Client commands shorter than eight characters could not run.** The command-shape tests indexed from `command.size() - 8`, which underflows, so `substr` threw and the catch-all turned it into `Exception processing command line: basic_string::substr…`. Affected `cpu`, `run`, and even `check_x` — long enough for `check_` but not for the `_forward` suffix test that runs first. Three sites, in every module built on the shared client machinery.

**The settings diff reported already-saved changes as pending.** `save()` wrote the values but never dropped the pending markers, so an edit stayed "pending" for the process lifetime — and changed shape from `added` to `modified` with `old_value == new_value`.

**`has_key()` ignored staged deletions**, contradicting every other read path.

**`house_keeping()` dereferenced the settings instance without a null check**, on a path the scheduler drives on a timer.

**`nscp settings --show --path X` with no `--key`** printed nothing and exited 0 — its guard tested `vm.count()` on options carrying `default_value("")`, so it could never fire.

Plus two smaller ones from the same root cause (the well-known `host` key goes to a typed field, not the data map): NRDP's trace line never named a sender, and SMTP announced itself to every mail server as `localhost` instead of the agent's hostname.

## Test fixes

Three tests were broken rather than the code:

- `check_battery` searched perfdata for a label containing "charge", but the counter is named after the battery (`BAT1`), so **every machine with a battery** took the no-battery branch and failed.
- `settings_http_test` dialled `127.0.0.1:1` expecting a refusal; WSL2 swallows connects to unbound fixed ports, so three tests sat in the TCP connect timeout — 405 seconds, passing.
- `lua_nrpe_relay_test` needed write access to `/usr/local/lib/nsclient/security`, so NRPEServer refused to load for anyone running ctest as themselves.

And one flake: every REST suite starts its agent on the same port 8443 and they run back to back, so a suite could inherit the previous one's agent — `waitForPort` only asks whether *someone* is accepting. That produced runs where the first few checks answered UNKNOWN (wrong agent) and the rest got ECONNREFUSED.

## Coverage added

| area | before | after |
|---|---|---|
| `libs/settings_manager/settings_handler_impl.cpp` | 24% | 94% |
| `service/settings_query_handler.cpp` | 23% | 70% |
| `modules/SyslogClient/syslog_client.hpp` | 0% | 95% |
| `modules/NRDPClient/nrdp_client.hpp` | 0% | 91% |
| `modules/CheckDisk/check_{share,uncpath,storagepool,shadowcopy}` | 0% | 84–98% |
| `modules/IcingaClient/icinga_client.hpp` | 0% | 69% |
| `include/client/command_line_parser.cpp` | 37% | 64% |
| `service/cli_parser.cpp` (integration) | 48% | 65% |
| `service/plugins/zip_plugin.cpp` | 0% | 53% |

Two of those were 0% only because their tests were built nowhere: the CheckDisk four live in a target gated `if(WIN32)` though only their data acquisition is Windows-specific, and `icinga_test.cpp` was in a target that omitted half the module.

## Left alone deliberately, with characterization tests

- `settings_ini::validate()` cannot report an unregistered key — it catches an exception the production core never throws. Fixing it starts reporting every key of every module that is not loaded, which is an output decision.
- `NSCAClient`, `IcingaClient` and `CollectdClient` read `timeout` from the data map, where `destination_container` never puts it, so a configured timeout is silently ignored. The fix has to choose a default: the container says 10, these clients document 30.
- `INISettings::has_real_path()` treats an existing empty section as missing.

## Verification

Unit 69/69. Integration 499 passed, 0 failed (177 skipped — docker suites). Both on Linux; the Windows-only paths are unchanged but unverified here.

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -18,6 +18,11 @@ jobs:
     uses: ./.github/workflows/tests-sanitizers.yml
     with:
       version: ${{ needs.get-version.outputs.version }}
+  coverage:
+    needs: get-version
+    uses: ./.github/workflows/tests-coverage.yml
+    with:
+      version: ${{ needs.get-version.outputs.version }}
   build-server-x64:
     needs: [get-version, create-test-bundle]
     uses: ./.github/workflows/build-windows.yml
--- a/.github/workflows/tests-coverage.yml
+++ b/.github/workflows/tests-coverage.yml
@@ -1,0 +1,151 @@
+name: Coverage (unit + integration)
+
+# Builds NSClient++ on Linux with gcov instrumentation (NSCP_COVERAGE=ON), runs
+# both the C++ unit tests (ctest) and the tests/ jest integration suite against
+# the same instrumented tree, and publishes an HTML + Cobertura coverage report
+# as a build artifact.
+#
+# The integration numbers are the interesting half: the jest harness spawns the
+# instrumented `nscp` and lets it dlopen the instrumented modules, so the report
+# covers real command dispatch and REST argument handling, not just the pure
+# logic the unit tests reach.
+#
+# Local reproduction (one command):
+#   tools/coverage/run.sh
+# or a single suite:
+#   SUITES=unit tools/coverage/run.sh
+#
+# This job is informational - it does not gate the build. Add
+# `--fail-under-line N` to the final gcovr call in tools/coverage/run.sh if you
+# ever want it to.
+
+permissions:
+  contents: read
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        type: string
+        description: 'Version to build'
+        required: true
+        default: '0.0.0'
+  # Also runnable on its own, so a coverage report can be pulled for a branch
+  # without waiting for a full main build.
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: 'Version to build'
+        required: false
+        default: '0.0.0'
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:24.04
+
+    steps:
+    - name: Install dependencies
+      run: |
+        apt-get update
+        DEBIAN_FRONTEND=noninteractive apt-get install -y \
+          git curl unzip cmake build-essential python3-pip pkg-config \
+          libboost-all-dev libgtest-dev libgmock-dev \
+          protobuf-compiler libprotobuf-dev \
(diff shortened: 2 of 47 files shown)


Fantu:
not related to this commit but after I saw remove of dead code a small remember of remove rust/nscp_client that was moved to other repo

mickem:
> not related to this commit but after I saw remove of dead code a small remember of remove rust/nscp_client that was moved to other repo

Not sure I understand what you mean?

Fantu:
rust/nscp_client was not moved to https://github.com/mickem/check_nsclient ?

mickem:
> > not related to this commit but after I saw remove of dead code a small remember of remove rust/nscp_client that was moved to other repo

Yes, thats the reference I understood... but not what you meant with the comment, sorry.. it is Saturday so am bit sleepy 😆

PR #1431, opened 2026-08-22
Configurable number rendering for check messages

### LINKED ISSUE #1428: feature request NRPEServer: numbers formatting (decimal places and separators)
If this feature already exists somehow, I just haven't found it:

Using NRPEServer with check_nrpe, I would like to be able to customize the number formats for the output of commands like check_drivesize or check_memory.

In my opinion, byte values with three decimal places are a bit unnecessary and hard to read:
example: `OK: Drive C: - 25.36% (used: 25.191GB, free: 74.136GB, total: 99.327GB) `

Maybe this could be made configurable, e.g. to 1 or 2 decimal places (for byte values and optionally for percent values as well)?

It would also be great if you could configure the decimal separator.
In Europe we are more used to commas than periods ;)

Maybe some parameter to switch between US and European style?  
Or perhaps even a variable to enter a separation character freely?

I'm still a little used to the output of the old check_nt protocol, which, in my opinion, does it perfectly out of the box:
example: `C: - total: 99,33 Gb - used: 25,19 Gb (25%) - free 74,14 Gb (75%) `
(don't know it the commas here are hardcoded or if it reads my system's regional settings)

mickem:
Fixes #1428.

`check_drivesize` reports `140.293GB/0.983TB used`: two units and six decimals for one disk, with no way to change either. The issue asked for fewer decimals and a European decimal comma. The mixed units in that one line turn out to be the larger readability problem, and neither of the two checks the issue names (`check_drivesize`, `check_memory`) could even reach the existing `format_bytes()` workaround — `register_format_functions()` was never called from them.

## The options

Four options, added next to `list-separator` in `add_misc_options()`, so every filter check has them and they fold out of the per-command reference onto the shared common-options page:

| Option | Effect | Default |
|---|---|---|
| `decimals` | Exactly N decimals (`2` → `25.20`) | `-1` (up to 3, trailing zeros stripped) |
| `byte-unit` | Pin every byte value to one unit, `B`…`EB` | auto, per value |
| `decimal-separator` | Radix character | `.` |
| `thousands-separator` | Digit grouping for the integer part | none |

```
check_drivesize drive=/ show-all=true
OK /: 141.085GB/0.983TB used

check_drivesize drive=/ show-all=true decimals=2 byte-unit=GB
OK /: 141.09GB/1006.85GB used

check_drivesize drive=/ show-all=true decimals=2 byte-unit=GB decimal-separator=, thousands-separator=.
OK /: 141,09GB/1.006,85GB used
```

The format lives on the evaluation context — the one thing every keyword getter and every format function is already handed — so the check's options set it once before the first row is evaluated. Defaults are unchanged, so an installation that sets none of them renders byte for byte what it rendered before.

**Two constraints held throughout:**

- It reaches the *message* only. Performance data is built from the raw values and keeps its full precision and its `.` radix: `'/ used'=141.03804GB;805.48314;…` even with a decimal comma configured. An integration case asserts this.
- Threshold parsing is untouched. `warning=used>1.5g` means the same thing with `decimal-separator=,` in force.

Real-time filters take the same values as `decimals`, `byte unit`, `decimal separator` and `thousands separator` settings keys, inheritable from the default template.

## Formatter bugs found on the way

- `format_byte_units(v, unit)` compared the unit case-sensitively against an uppercase table, so `format_bytes(used, 'gb')` rendered `1.27055e-10` while `convert_bytes(used, 'gb')` — which upper-cases — worked. Units are now case insensitive everywhere.
- A unit that matched nothing fell out of the loop having divided seven times, so a typo rendered `value/1024^7`. An unknown unit is now reported.
- `convert_to_byte_units` had the same fallthrough, so a misspelled `unit:` in `perf-config` silently scaled the metric by 1024⁻⁷. It now leaves the value alone. **A graph that has been flat near zero will jump to its real magnitude on upgrade.**

Reporting the typo needed one more fix: a function that failed while a template rendered called `context->error()`, and nothing on the rendering path read it — the placeholder just came out empty. `modern_filters` now drains those into the error handler the checks already report through, so the check returns UNKNOWN with the reason instead of a quietly incomplete message.

## Functions

`format_bytes()` gained an optional `decimals` argument, and `format_number(value, decimals)` joins it for percentages and other non-byte values. Both follow the options above unless given an explicit argument.

`register_format_functions()` is now called from every check carrying byte-valued keywords that lacked it — `check_drivesize` and `check_memory` among them.

## Testing

- **Unit:** new `str/number_format_test.cpp` for the rendering primitives; `format_test.cpp` updated for the three fixed contracts; `cli_helper_test.cpp` covers the option plumbing, the valued (`--decimals=1`) form REST uses, and rejection of a bad `byte-unit`/`decimals`.
- **Integration:** 8 new cases in `checkdisk-unix.test.ts` driving the options over the one-shot client-query path (same `k=v` single-token parsing as REST), including the locale-neutral perfdata assertion and the `format_bytes`/`format_number` templates.
- Full unit suite passes locally except `lua_log_file_test`, which fails on this machine because `os.tmpname()` hands the Lua script a `/tmp` path the sandbox won't let it write — unrelated, and it fails the same way on a clean tree.
- The REST-based suites (`checkdisk-commands`, `checksystem-commands`, `checknet-commands`) could not run locally: no build dir in this checkout has `libWEBServer.so`, so nothing can bind :8443. They were unrunnable before this change too and need CI.

## Needs a Windows build before merge

The Windows-only sources (`check_drive_win.cpp`, `check_process.cpp`, `filter.hpp`, `check_kernel_memory`, `check_cpu_frequency`, `check_hardware`, `check_network`, `check_pdh`, `check_uncpath`, `win/processes.hpp`) follow the same mechanical pattern as their Linux counterparts — human getters take an `evaluation_context`, registrations move to `add_human_string_context` / `add_string_var_w_context` — but they are not compiled on this machine.

## Docs

`docs/reference/*.yaml` and the generated Markdown were regenerated, keeping only what this change causes. A "Number formatting" section was added to `concepts/checks.md`, and `setup/upgrading.md` carries four entries covering the behaviour changes above.

### DIFF
--- a/docs/docs/concepts/checks.md
+++ b/docs/docs/concepts/checks.md
@@ -428,6 +428,47 @@ long output, shown as a block — which is the point of the exercise.
     a long multi-line result is cut off sooner than the single-line form would be.
 <!-- @formatter:on -->
 
+### Number formatting
+
+By default every byte value scales on its own and carries up to three decimals, which is how a
+single line ends up reading `141.085GB/0.983TB` — two units, six decimals, one disk. Four options
+change that, and they are available on every filter check:
+
+| Option                 | Effect                                                        | Default          |
+|------------------------|---------------------------------------------------------------|------------------|
+| `decimals`             | Decimals to render, exactly (`2` → `25.20`)                   | `-1` (up to 3)   |
+| `byte-unit`            | Pin every byte value to one unit: `B`…`EB`                    | auto per value   |
+| `decimal-separator`    | Radix character — `,` for the European rendering              | `.`              |
+| `thousands-separator`  | Digit grouping for the integer part                           | none             |
+
+```shell
+check_drivesize drive=/ show-all=true
+OK /: 141.085GB/0.983TB used
+
+check_drivesize drive=/ show-all=true decimals=2 byte-unit=GB
+OK /: 141.09GB/1006.85GB used
+
+check_drivesize drive=/ show-all=true decimals=2 byte-unit=GB decimal-separator=, thousands-separator=.
+OK /: 141,09GB/1.006,85GB used
+```
+
+They apply to the byte and percentage keywords and to the
+[functions](#6-functions-transforming-values) below.
+
+<!-- @formatter:off -->
+!!! note "Only the message changes"
+    Performance data is generated from the raw values, so it keeps its full precision and its `.`
+    radix character whatever you set here — graphs and time series are unaffected. The numbers you
+    write in a `filter`, `warning` or `critical` are parsed by the filter language and always use
+    `.` as well: `warning=used > 1.5g` means the same thing with `decimal-separator=,` in force.
+
+!!! tip "Setting it once"
+    These are ordinary check options, so an alias fixes them for a whole command:
+    `[/settings/external scripts/alias] alias_disk = check_drivesize decimals=2 byte-unit=GB`.
+    Real-time filters take the same values as the `decimals`, `byte unit`, `decimal separator` and
+    `thousands separator` settings keys, which inherit from the default template.
+<!-- @formatter:on -->
+
 ---
 
 ## 6. Functions — Transforming Values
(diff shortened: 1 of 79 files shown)


mickem:
Superseded by #1434, which carries these three commits reworked onto latest main (resolving the conflict with the cross-type filtering changes from #1433) and adds an end-to-end rendering test suite — the output-side companion of `filter_matrix_test` — covering the message rendering both with and without the new formatting options. Closing in its favor.

---

PR #1435, opened 2026-08-24
feature: release 0.17.0

mickem:
Stamp the pending documentation entries with the release they ship in. "## next" in the upgrade guide and the "Fixed in: next" hardening notice both become 0.17.0.

Drop the "runaway decimals" hardening notice from the security notices page - the decimals cap is documented as an ordinary behaviour change in the upgrade guide, which no longer links to the removed anchor.

Move the summary-keyword rename entry from 0.16.4 to 0.17.0: it landed after the 0.16.4 tag was cut, so it first ships here. For the same reason, correct two "in 0.17" notes that describe changes already released - settings query parameters (0.16.1) and the log file moving into ${log-path} (0.16.2).

Assisted-by: Claude Code:claude-opus-5


dneuhaeuser:
thank you very much!!
that works perfectly!

PR #1453, opened 2026-08-29
ElasticClient: add authentication, TLS verification, and modernize

mickem:
## Summary

Modernize the `ElasticClient` module to support authentication, verify HTTPS server certificates by default, add network timeouts, fix silent data loss in bulk submissions, and drop the legacy `_type` parameter that Elasticsearch 8+ rejects.

## Key Changes

- **Authentication support**: New `user`/`password` (basic auth) and `api key` settings add the `Authorization` header to submissions. Both are stored as password-typed settings and never logged.

- **HTTPS certificate verification**: The module previously hardcoded TLS verification to `none`. Now defaults to `peer` against the platform CA bundle, with new `tls version`, `verify mode`, and `ca` settings matching other HTTP client modules.

- **Network timeout**: New `timeout` setting (default 30 seconds) bounds each connect/read/write operation, preventing stalled servers from wedging the agent indefinitely.

- **Per-document IDs**: Every document in a bulk request now gets its own random UUID `_id`. Previously all documents in a batch shared the same ID, causing silent data loss (all but the last document would be overwritten).

- **Drop legacy `_type` parameter**: The `event type`, `metrics type`, and `nsclient log type` settings now default to empty strings. The `_type` parameter is only emitted when explicitly configured, since Elasticsearch 8+ rejects it. Operators running 6.x or older can restore the old behavior by setting these values.

- **Defensive error parsing**: Response bodies are now parsed without assuming a shape, and error text included in logs is truncated to prevent log injection.

- **Extract bulk helpers**: Moved `parse_index()`, `build_payload()`, and error extraction logic into a new `elastic_bulk.hpp` header with comprehensive unit tests (`ElasticClient_test.cpp`).

- **Integration test**: New `elastic-submit.test.ts` verifies the wire contract: x-ndjson framing, basic auth header, per-document IDs, `%(date)` expansion, and absence of `_type` by default.

## Implementation Details

- The `send_to_elastic()` function is now a member method with access to configuration (credentials, TLS settings, timeout).
- UUID generation moved to `build_payload()` so each document gets a distinct ID.
- Error extraction handles non-JSON responses, missing fields, and hostile shapes without throwing.
- The integration test uses an in-process HTTP server (no Docker) to capture and verify the exact request format.

https://claude.ai/code/session_01S83Vko1KQiWogAm8AVuhmA


mickem:
Addressed the review findings in 3e9aaee. Five of the six led to changes; one I did not apply, with reasoning below.

**Fixed**

- **Log injection via the error text** — genuinely incomplete before: truncating the body left any CR/LF in a hostile `reason` free to forge extra log lines. Every string `extract_errors()` returns now goes through `str::format::strip_ctrl_chars` (the helper already used for the same purpose in the NSCA path), covered by a new unit test asserting no CR/LF survives for both a JSON `reason` and a non-JSON body.
- **Data race on `started`** — now `std::atomic<bool>`. The event/metrics/log callbacks read it on core threads while `unloadModule()` writes it.
- **Negative `timeout` underflow** — confirmed real and worse than cosmetic: `str::stox<unsigned int>("-5")` yields 4294967291, so `timeout = -5` wrapped to ~4.29 billion seconds and silently restored the unbounded wait this setting exists to prevent. `timeout` is now `int`, registered with `int_key`, and a negative value is rejected at load (logged, falls back to 30).
- **`parse_index()` hoisted** out of the per-document loop — same result for every action line in one request.
- **Integration-test framing assertion** added before indexing into the parsed lines.

**Not applied**

- **`${Boost_SYSTEM_LIBRARY}` on the test target.** Boost.System has been header-only since 1.69 and this project builds against 1.86, so no link input is required — and no other `CMakeLists.txt` in the tree links it, so adding it here would be inconsistent rather than defensive. The test target builds and links clean on Linux/GCC; if a toolchain ever does need it, the right fix is repo-wide rather than in this one module.

Verified before pushing: full rebuild clean, `ctest` green (only the pre-existing `ZipPluginTest` sandbox failure, unrelated to this change), and the `elastic-submit` integration suite passes against the built agent.

---

PR #1455, opened 2026-08-29
fix(core): load modules enabled since the last reload

mickem:
Follow-up to #1454, which worked around this in the provisioning scripts. This fixes the underlying behaviour in the agent.

## The bug

Reloading re-scans `[/modules]` to decide what should be running, but reads that list through the settings store **as it was cached at load time**. The cache holds the child instances built from `[/includes]`, so a module enabled since then — by an edit plus a reload, or by a fleet bundle that turns one on — was never in the list `find_all_active_plugins()` walked, and so was never loaded.

Nothing reported a problem, which is what makes it nasty. Every diagnostic you would naturally reach for looked right:

- `fleet.ini` on disk said the module was enabled
- `nscp settings --list --path /modules` agreed (that path reads through a fresh store)
- the fleet host reported itself **in sync**

Only the loaded libraries told the truth:

```
before restart: libCheckDisk.so libCheckHelpers.so libCheckSystem.so
after restart:  libCheckDisk.so libCheckHelpers.so libCheckSystem.so
                libNRDPClient.so libScheduler.so libWEBServer.so
```

Found while building the turn-key Nagios flow in #1454: a bundle enabling `NRDPClient` and `Scheduler` applied cleanly, the hosts went in sync, and not one passive result was ever submitted.

## The fix

`reloadPlugins()` clears the settings cache before re-scanning. The backend then re-reads its file and rebuilds the `[/includes]` children — `fleet.ini` among them — so the scan sees the configuration as it is on disk right now.

`clear_cache()` is not a new operation on this path: `do_reload("settings")` already did exactly this, and the ini backend's `real_clear_cache()` re-reads the file and repopulates the children, so the includes are rebuilt rather than lost (asserted in one of the tests). A module already running is recognised as a duplicate by `only_load_module()` and left alone, so this only ever adds.

**Not fixed, and now noted in the code:** a module *disabled* since the last load keeps running until the service restarts. That needs unloading a live plugin — a different problem, and a riskier one.

## Tests

Two cases on the settings manager, which is where the staleness actually lives (`reloadPlugins()` itself needs a whole `NSClientT`):

- a module added to an included file is invisible until the cache is cleared — the `EXPECT_FALSE` is the bug, so this fails to protect nothing if the caching ever changes
- clearing the cache does not lose the includes

`settings_manager_test` 86/86 pass, `plugins_test` 118/118 pass, and `nscp.exe` builds. `service_test` has 7 pre-existing `AclTest` failures on my machine; I confirmed they are identical with this change stashed, so they are unrelated (Windows ACL tests on a non-English host).

## Docs

An `upgrading.md` entry under 0.17.2, since this changes observable reload behaviour — operators who worked around it by restarting the service can stop.

https://claude.ai/code/session_012fAQW2wZxkk9QbCnX6zEQR

### DIFF
--- a/libs/settings_manager/settings_manager_impl_test.cpp
+++ b/libs/settings_manager/settings_manager_impl_test.cpp
@@ -1052,6 +1052,92 @@ TEST_F(SettingsContextTest, AContextWithoutAPlaceholderIsUnaffected) {
   EXPECT_FALSE(impl_->context_exists(ini_context(dir.path() / "${host}-plain.ini")));
 }
 
+// --- what a reload has to do before it can see a new module ----------------
+//
+// NSClientT::reloadPlugins() re-scans /modules to decide what to load. The
+// list it reads is served partly from the child instances built when the
+// configuration was last loaded, so a module enabled in an *included* file
+// since then - fleet.ini, written by the fleet sync, is the case that matters
+// - stays invisible until that child re-reads its file. That is why the reload
+// refreshes the includes first: without it a fleet bundle enabling a module
+// applied cleanly, the host reported itself in sync, and the module was only
+// really loaded the next time the service happened to restart.
+//
+// It refreshes the *children* rather than clearing the whole store, and the
+// last test here is why: clearing the store also discards configuration held
+// in memory and never saved, which is exactly how `nscp unit` and
+// `nscp client` set themselves up before asking for a reload.
+
+class SettingsIncludeReloadTest : public SettingsHandlerTest {
+ protected:
+  settings_test::temp_dir dir_;
+  boost::filesystem::path main_ini_;
+  boost::filesystem::path included_ini_;
+  std::unique_ptr<passthrough_provider> provider_;
+
+  void SetUp() override {
+    provider_ = std::make_unique<passthrough_provider>();
+    impl_ = std::make_unique<settings_manager::NSCSettingsImpl>(provider_.get());
+    main_ini_ = dir_.file("nsclient.ini");
+    included_ini_ = dir_.file("fleet.ini");
+    write_included("CheckSystem = enabled\n");
+    settings_test::write_file(main_ini_, "[/includes]\nfleet = " + ini_context(included_ini_) + "\n");
+    impl_->set_instance("master", ini_context(main_ini_));
+  }
+
+  void write_included(const std::string &modules) { settings_test::write_file(included_ini_, "[/modules]\n" + modules); }
+
+  settings::settings_interface::string_list modules() { return impl_->get()->get_keys("/modules"); }
+
+  // What NSClientT::reloadPlugins() does before re-scanning /modules.
+  void refresh_includes() {
+    for (const settings::instance_ptr &child : impl_->get()->get_children()) {
+      if (child) child->clear_cache();
+    }
+  }
+
+  static bool has(const settings::settings_interface::string_list &keys, const std::string &key) {
+    return std::find(keys.begin(), keys.end(), key) != keys.end();
+  }
+};
+
+TEST_F(SettingsIncludeReloadTest, ModulesFromAnIncludedFileAreVisible) {
+  EXPECT_TRUE(has(modules(), "CheckSystem"));
+}
+
+TEST_F(SettingsIncludeReloadTest, AModuleAddedToAnIncludeAppearsOnlyAfterTheIncludesAreRefreshed) {
+  ASSERT_TRUE(has(modules(), "CheckSystem"));
+
+  write_included("CheckSystem = enabled\nNRDPClient = enabled\n");
+
+  // Served from the child built at load time, so the new module is not here
+  // yet. A reload that skipped the refresh would scan this same stale list and
+  // conclude there was nothing new to load.
+  EXPECT_FALSE(has(modules(), "NRDPClient"));
+
+  refresh_includes();
+
+  EXPECT_TRUE(has(modules(), "NRDPClient"));
+  EXPECT_TRUE(has(modules(), "CheckSystem")) << "refreshing must not lose what the include already had";
+}
+
+TEST_F(SettingsIncludeReloadTest, RefreshingTheIncludesKeepsConfigurationSetInMemory) {
+  // The reason the reload refreshes the children instead of clearing the whole
+  // store. `nscp unit` and `nscp client` enable their modules in memory and
+  // then ask for a reload to apply them; nothing is ever saved to disk. A
+  // reload that cleared the store would drop this and leave them with no
+  // modules at all - which is precisely what a first attempt at the fix did.
+  impl_->get()->set_string("/modules", "PythonScript", "enabled");
+  ASSERT_TRUE(has(modules(), "PythonScript"));
+
+  refresh_includes();
+
+  EXPECT_TRUE(has(modules(), "PythonScript")) << "an unsaved in-memory module must survive the refresh";
+  EXPECT_EQ(impl_->get()->get_string("/modules", "PythonScript", ""), "enabled");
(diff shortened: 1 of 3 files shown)


mickem:
Force-pushed a corrected fix. The first attempt was wrong and broke every acceptance test — thanks CI.

**What went wrong:** I cleared the *whole* settings store before re-scanning `/modules`. That also discards configuration set in memory and never saved — which is exactly how `nscp unit` and `nscp client` work (`cli_parser.cpp`: `run_pre` configures in memory, then `run_reload` → `reload("instant,service")` to apply it). With the store emptied, no module was enabled any more, hence `No handler for command: py_unittest_add_case` with `commands {}`.

**The corrected fix** refreshes only the `[/includes]` children, so each re-reads its own file while in-memory configuration is left untouched. That still covers the real bug (the fleet sync rewrites `fleet.ini`, an include), and values read from a child aren't memoised by the parent, so settings changes in an include take effect too.

Narrower than the original claim, and the upgrade note now says so: `[/modules]` in the main `nsclient.ini` is still only read at startup.

**Verified locally**, reproducing the CI failure first:

| | before fix | broken attempt | corrected |
| --- | --- | --- | --- |
| `nscp unit --language lua --script test_nrpe.lua` | 81 cases, exit 0 | `Failed to find command: lua_unittest`, exit 3 | 81 cases, exit 0 |
| `nscp unit --language python --script test_nsca --case none` | — | (the CI failure) | 644 tests, exit 0 |

A third unit test now guards this specific regression — it asserts an unsaved in-memory module survives the refresh, and I confirmed it **fails** against the broken approach rather than trusting that it would.

service/NSClient++.cpp
```diff
@@ -616,10 +616,30 @@ void NSClientT::unloadPlugins() {
...
+  // while the module only really appeared at the next service restart.
+  //
+  // Only the children are refreshed, and deliberately so. Clearing the whole
+  // store would also throw away configuration that was set in memory and never
+  // saved - which is precisely how `nscp unit` and `nscp client` work: they
+  // configure the agent in memory and then reload to apply it. Dropping that
+  // leaves them with no modules at all.
+  for (const settings::instance_ptr &child : settings_manager::get_settings()->get_children()) {
+    if (child) child->clear_cache();
+  }
```
mickem, line 649: `clear_cache()` can throw: the ini backend's `real_clear_cache()` → `load_data()` raises `settings_exception` via `throw_SI_error` when the include's file can't be read at reload time (deleted, transient permission problem — plausible mid-rewrite by the fleet sync). Thrown here, it aborts `reloadPlugins()` before the re-scan (caught and logged in `do_reload`), and the failed child is left empty — its caches and children were cleared before the load failed — so subsequent `/modules` scans lose that include's modules entirely until a later reload succeeds.

A per-child try/catch (log the bad include, keep refreshing the rest) would contain the damage to the one broken file. The same exposure pre-exists in `do_reload("settings")`'s whole-store `clear_cache()`, so this is a hardening suggestion, not a blocker.

---

libs/settings_manager/settings_manager_impl_test.cpp
```diff
@@ -1052,6 +1052,92 @@ TEST_F(SettingsContextTest, AContextWithoutAPlaceholderIsUnaffected) {
...
+// last test here is why: clearing the store also discards configuration held
+// in memory and never saved, which is exactly how `nscp unit` and
+// `nscp client` set themselves up before asking for a reload.
+
+class SettingsIncludeReloadTest : public SettingsHandlerTest {
+ protected:
+  settings_test::temp_dir dir_;
+  boost::filesystem::path main_ini_;
+  boost::filesystem::path included_ini_;
+  std::unique_ptr<passthrough_provider> provider_;
```
mickem: Nit: this `provider_` shadows `SettingsHandlerTest::provider_` (a `unique_ptr<recording_provider>` that stays null here because `SetUp` is overridden). It works, but `SettingsContextTest` sidestepped the same situation by using a distinct name (`path_provider_`); doing the same here would keep the fixtures consistent and avoid a `-Wshadow-field` on compilers that flag it.

---

mickem (commented):
Reviewed the fix against the settings-store internals; the change is correct, minimal, and the "only refresh the children" scoping holds up under scrutiny. Everything the PR description claims checks out in the code:

**Verified in the source**

- **The refresh actually re-reads the file and keeps the includes.** `settings_interface_impl::clear_cache()` clears the child's caches *and* its `children_`, then the ini backend's `real_clear_cache()` re-runs `load_data()`, which re-reads the file and re-adds `[/includes]` children as fresh instances (`create_instance` never returns a cached object). Nested includes are therefore refreshed too, and the pre-cleared `children_` means no duplication.
- **Clearing only the children is sufficient, not just safe.** The parent never memoises values it resolved from a child — `getter()` only populates `settings_cache_` from its own `get_real_*`; child hits are returned directly — and `get_keys()`/`get_sections()` aggregate the children live on every call. So a value *changed* in an include is picked up as well as a key added, with nothing stale left in the parent.
- **Changed values really do overwrite.** With multikey off, `CSimpleIni::AddEntry` replaces the value for an existing key on re-load, so the docs claim "settings changed in them take effect" is accurate.
- **Modules already running are left alone.** `only_load_module()` returns the `find_duplicate()` hit without loading, exactly as the comment says.
- **The `nscp unit`/`nscp client` regression guard is the right test.** Unsaved values set through the parent land in the parent's `settings_cache_` (the setter only routes to a child when the key already exists there), which the loop doesn't touch.

**Minor observations, none blocking**

1. Exception safety of the loop — inline comment on `NSClient++.cpp`.
2. One narrower in-memory case *is* dropped: an unsaved `set_string` on a key that already exists in an included file is routed into that child's cache by `setter()`'s `set_in_child` path, so the refresh reverts it to the on-disk value. That's arguably the right semantics for a reload (and `do_reload("settings")` already did the same), but it's a behaviour change relative to the old `reloadPlugins()`, worth being a conscious decision.
3. A key *deleted* from an include survives the refresh: `load_data()` calls `ini.LoadFile` on the existing `CSimpleIni` without a `Reset()`, and load merges rather than replaces. It's the same family as the documented "disabled module keeps running" limitation (and pre-existing), but note the upgrading entry's "settings changed in them take effect" holds for changed values, not removals.
4. `refresh_includes()` in the test re-implements the production loop rather than calling it — acknowledged in the comments as the price of `NSClientT`'s weight; if `reloadPlugins()` ever changes its refresh strategy the tests will keep passing while diverging, so worth remembering they're a specification of the loop, not coverage of it.
5. Fixture naming nit — inline comment on the test file.

The upgrading.md entry is accurate and appropriately scoped (per point 3, "modules disabled … keep running" could arguably also mention removed keys, but that's a footnote). CI was still part-way through at review time; everything completed so far is green, including DCO and both deb/rpm build+unit-test legs.

---

service/NSClient++.cpp
mickem: Fixed in b00391a: `clear_cache()` is now wrapped per child, so a throwing include is logged with its context and costs only its own content — the remaining includes still refresh and the reload proceeds to the re-scan.

One precision on the scenario: a *deleted* include never threw — `load_data()` checks `file_exists()` and just comes back empty — so the catch matters for a file that exists but cannot be read (permissions, transient I/O). The code comment notes both cases.

---

libs/settings_manager/settings_manager_impl_test.cpp
mickem: Fixed in b00391a: renamed to `include_provider_`, with a comment noting the base fixture's `provider_` stays null here because `SetUp` is overridden — matching `SettingsContextTest`'s `path_provider_` pattern.

---

PR #1457, opened 2026-08-29
Fix silently ignored client target settings; raise unit coverage for sub-50% files

mickem:
## Summary

A unit-coverage sweep of the gcovr reports (`coverage/unit.json` + `coverage/integration.json`) surfaced two real bugs, both the same defect class: a client target object's `read()` building a settings registry and then never calling `register_all()`/`notify()`, so the configured values were silently discarded.

### Bug fixes (separate commits)

- **CheckMKClient** 🔒: the SSL keys (`use ssl`, `certificate`, `verify mode`, `ca`, …) on a `[/settings/check_mk/client/targets/…]` section were never applied — the client always connected in **plaintext** regardless of configuration, and the keys were missing from the generated docs. Security notice added under *Hardening changes (no CVE)*, with the matching 🔒 `upgrading.md` entry (a target carrying an old `use ssl = true` starts negotiating TLS on upgrade).
- **SyslogClient**: `severity`, `facility`, the templates and per-status severities on a target section were never applied; the constructor defaults always won. `upgrading.md` entry added since previously inert configuration now takes effect.

Both fixes are pinned by new handler test suites (including the valued `--ssl=false` form REST-style dispatch depends on).

### Test coverage batch

Every source file under 50% combined line coverage that can be exercised deterministically on this platform now has unit tests (~270 new test cases; suite grows 69 → 77 binaries). Highlights: check_mk wire-protocol round-trips, `pid_file` O_EXCL/O_NOFOLLOW behaviours, `execute_process_unix` timeout-kill/no-shell guarantees, NSCA-ng TLS handshake-failure paths, onboarding `chown_subtree` via a user-namespace fake-root harness, `simple_file_logger` rotation, and a temp-sysfs fixture for cpu-frequency parsing (small no-behaviour-change refactor to accept a base path).

Measured after (gcovr, previously <50%): `nsca_ng_handler` 100%, `nscp_client.hpp` 94%, `simple_file_logger` 89%, `onboarding/state` 89%, `engine_impl.hpp` 97%, CheckSystemUnix network/service/cpu-frequency 100%, `settings_proxy` 100%, and more.

Intentionally not covered: interpreter wrappers (Python/Lua `script_wrapper`, `extscr_cli`), paths needing live servers/backends (post-handshake NSCA-ng/NSCP exchanges, WEBServer controllers, MySQL/Docker/Elastic/Op5), and `pinger.hpp` past socket-open (needs CAP_NET_RAW; an end-to-end test runs when privileged and skips otherwise).

## Test plan

- [x] Full `ctest` in a configured build dir: **77/77 pass**
- [x] Coverage deltas verified with gcovr in the instrumented coverage build for the larger suites


mickem:
Thanks — all review points addressed, plus the two CI failures.

**Missing includes** (3 comments): added `<algorithm>` to `client_protocol_test.cpp` and `syslog_handler_test.cpp` (for `std::copy`/`std::find`) and `<ctime>` to `execute_process_unix_test.cpp` (for `time_t`/`time()`) in 4d3cd997.

**Removed NSCA-NG upgrade/security notes** (3 comments): good catch, this was a real regression and not intentional. Rebasing this branch onto main resolved the docs conflict using a resolution that predated #1461, which silently dropped the NSCA-NG cert-mode entries that had landed on main in the meantime. Restored in 9e508612 by taking main's current pages and re-applying the check_mk/syslog entries on top, so both sets coexist under `## Unreleased`.

**Windows link failure** (`check_mk_handler_impl_test`, LNK2019 on `nscapi::logging::parse`): lua_nscp's `core_wrapper::log()` needs that symbol and the Windows plugin_api DLL doesn't export it — `nscapi_helper.cpp` is now compiled into the target (9299a342), same workaround the other test targets use.

**`service_logger_test` failure**: this one turned out to be a genuine Windows-only bug rather than a flaky test. `simple_file_logger`'s constructor prepended `base_path()` unconditionally, while `asynch_configure()` only does so for a name without a path separator. On Windows `base_path()` is the executable's directory, so passing an absolute path produced `D:\a\nscp\nscp\tmp\nscpC:/Users/RUNNER~1/.../output.log` and the log was never written. Fixed in d23321c7 by applying the same bare-name rule in the constructor; production is unaffected since the only caller passes `nsclient.log`. The bare-name settings test now also runs inside its own temp dir rather than depending on the suite's working directory being writable.

Full suite is green locally (77/77).

PR #1467, opened 2026-08-30
GraphiteClient: fix metric injection and enforce submission timeout

mickem:
## Summary

This PR fixes two security and availability issues in GraphiteClient:

1. **Metric-line injection vulnerability**: The status path was only scrubbed for spaces, allowing check aliases from remote submitters (NSCA, forwarded results) to inject newlines or semicolons that corrupt metrics or inject fake ones. The status path now receives the same comprehensive scrub as the perf path.

2. **Broken timeout enforcement**: The `timeout` setting had no effect — the configured value never reached the connection (the lookup read the wrong data map), and the connection made no use of it anyway (resolution, connect, TLS handshake, and writes all ran synchronously with no deadline). A stalled carbon endpoint could hold the submitting thread indefinitely, blocking the recurring metrics flush.

## Key Changes

- **Status path scrubbing**: Extended `fix_graphite_string()` to replace tabs (which split carbon fields like spaces do) in addition to existing replacements. Applied this scrub to the status path, matching the perf path treatment.

- **Deadline-bounded I/O**: Introduced `detail::deadline_io` class that races all async operations (resolve, connect, TLS handshake, write) against a single budget deadline. The `timeout` setting (default 30s) now bounds the entire submission as one budget, following the SMTP client precedent.

- **Timeout configuration fix**: Changed `connection_data` to read `timeout` from the typed field (`target.timeout`) instead of the free-form data map, where the well-known key is never stored.

- **Removed retry reading**: The module always made exactly one attempt per submission; the `retry` setting is no longer read to avoid implying it is honored.

- **Comprehensive test coverage**: Added `graphite_client_test.cpp` with 320 lines of tests covering:
  - Metric-line injection prevention (hostile aliases and perf labels)
  - Benign submission rendering
  - Timeout enforcement under stalled TCP writes
  - TLS handshake timeout (when USE_SSL is available)

- **Documentation**: Updated security notices and upgrade guide to explain the fixes and operator impact.

## Implementation Details

- The `deadline_io` class uses `boost::asio::steady_timer` to enforce a shared budget across all operations, canceling outstanding I/O when the deadline expires.
- The scrub now replaces `\t`, `\r`, `\n`, `;`, NUL, spaces, and reserved characters with `_`, preventing both metric-line injection and carbon tag injection.
- Error messages distinguish between operator-configured timeouts and platform errors, improving diagnostics.
- The TLS shutdown path uses a short grace period (2s) capped by the remaining budget to avoid stretching successful submissions.

https://claude.ai/code/session_013HMntJaonUfGNGEdRooGFH

### DIFF
--- a/docs/docs/security/notices.md
+++ b/docs/docs/security/notices.md
@@ -171,7 +171,7 @@ legacy auth — and produced three hardening fixes on the client side:
   also no longer suggests `none` for self-signed certificates — the right
   configuration is `peer-cert` with `ca` pointing at the certificate.
 
-### Graphite client — metric-line injection scrub and a whole-submission timeout
+### Graphite client: metric-line injection scrub and a whole-submission timeout
 
 **Fixed in:** unreleased (first release after 0.18.0) · **Severity:** Low
 
--- a/docs/docs/security/notices.md
+++ b/docs/docs/security/notices.md
@@ -171,7 +171,7 @@ legacy auth — and produced three hardening fixes on the client side:
   also no longer suggests `none` for self-signed certificates — the right
   configuration is `peer-cert` with `ca` pointing at the certificate.
 
-### Graphite client — status path scrubbed against metric-line injection
+### Graphite client — metric-line injection scrub and a whole-submission timeout
 
 **Fixed in:** unreleased (first release after 0.18.0) · **Severity:** Low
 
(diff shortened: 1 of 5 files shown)


mickem:
Rebased onto `main`. Now **MERGEABLE** — was conflicting because the branch predated nine merged PRs.

**The injection-scrub commit dropped out as already applied.** #1465 merged the same change as `5f93c03d0`; I diffed the two patches for `graphite_client.hpp` and they are byte-identical, so `git rebase` correctly skipped it. What remains is the work that is *not* in main:

- `security: bound the whole Graphite submission by the configured timeout`
- `graphite: log a failed metrics submission instead of dropping it`

Worth restating why that matters: on current `main` the Graphite `timeout` is **read and then ignored** — `graphite_client.hpp` has zero deadline/timer code, so resolution, connect, the TLS handshake and every write run with no deadline and a stalled carbon endpoint holds the submitting thread. The lookup was broken too (`get_int_data("timeout", 30)` reads the free-form map, which the well-known `timeout` key never lands in, so the default always won).

The diff is now +366/-31 across 5 files, down from +629/-39 against the stale base.

## Conflict resolution

Three files needed hand-resolution; all in favour of what `main` has now:

- **`notices.md`** — kept main's Graphite notice body and *appended* the timeout paragraph, rather than replacing the entry. The heading now covers both topics, and I switched it from an em dash to a colon (`### Graphite client: …`) to match the other post-0.18.0 notices — an em dash slugifies to a double hyphen and left the upgrade note's link dangling. All nine `notices.md` anchors referenced from `upgrading.md` now resolve.
- **`upgrading.md`** — dropped the branch's own `## Next release after 0.18.0` heading and filed both Graphite entries under main's existing `## Unreleased` section.
- **`graphite_client_test.cpp`** — include-only conflict, took both sides.

## Verified

`graphite_client_test` builds and all **9 tests pass**, including the two new ones proving the deadline actually fires:

```
GraphiteTimeoutTest.stalled_receiver_fails_within_the_configured_timeout        (1097 ms)
GraphiteTimeoutTest.stalled_tls_handshake_fails_within_the_configured_timeout   (1019 ms)
GraphiteConnectionDataTest.timeout_is_read_and_retry_is_not                     OK
```

PR #1477, opened 2026-09-01
test: cover the posture check bodies and the WEBServer buffers

mickem:
A gcov run of the unit suite (tools/coverage/run.sh, SUITES=unit) put eleven
files at or near zero that nothing but a Windows machine could reach - or, in
the WEBServer case, that nothing reached at all. These are the tests for them:
they lift those files from 459 covered lines to 803.

CheckSecurity - the eight checks whose data acquisition is Windows-only
(check_nla, check_antivirus, check_bitlocker, check_secureboot, check_defender,
check_local_accounts, check_group_members, check_users) were entirely uncovered
off Windows, and only half-covered on it: the integration suite reaches the
"not supported on this platform" branch, which returns before a single row is
matched, and the existing unit tests stop at the filter_obj accessors. So the
part that decides whether a host is reported healthy or critical - the default
warning/critical expressions, the empty-state contract, the rendering and the
perf data - had no coverage anywhere.

Since only the gathering is platform-specific, the new test stands in for it,
the way check_windows_checks_unix_test.cpp already does for CheckDisk, and
hands each command rows from a fixture. That pins, among other things:

  * that a disabled Guest account is not a warning while an enabled one is, and
    that neither default fires on a disabled passwordless account - both
    defaults are qualified with `enabled = 1` on purpose;
  * that Defender's -1 ("unknown") ages never trip thresholds written as `> N`;
  * that check_secureboot's default still trips on a legacy BIOS boot, which is
    why `supported` is a keyword;
  * that check_users has no default alert at all, and that `count` is what an
    operator writes a threshold against;
  * that check_group_members' allow-list matches a bare name as well as
    DOMAIN\name, case-insensitively.

Two behaviours worth knowing are pinned as they are rather than changed. The
ok-syntax these checks ship never renders, because cli_helper drops it whenever
the top-syntax contains a list - which every one of them does; and perf data
follows the threshold expressions rather than the keyword registry, so
check_defender graphs only the signature age by default.

WEBServer - event_store (the bounded buffer behind /api/v2/events) was at 0%
and error_handler (the log buffer behind /api/v2/logs) at 20%; both are now
fully covered. The event tests pin the cap, the trim when the cap is lowered
while running, the zero-is-clamped-to-one rule, and that indices keep climbing
across a drain so a resuming stream client is never handed an index below one
it has already seen. The log tests pin the paging, whose three implementations
do not agree: the unfiltered branch slices [position, position+ipp), while the
level-filtered and "since" branches keep entries whose 1-based ordinal falls in
[position, position+ipp] - one earlier, and one more. The tests say so rather
than endorse it; changing it means changing the log viewer with it.

metrics_handler was already at 100% incidentally, so its tests add contract
rather than coverage: that the three representations are independent latches,
that each write replaces the whole snapshot, and that set_openmetrics() copies
rather than steals the caller's list.

Assisted-by: Claude Code:claude-opus-5
Signed-off-by: Michael Medin <michael@medin.name>

### DIFF
--- a/modules/CheckSecurity/CMakeLists.txt
+++ b/modules/CheckSecurity/CMakeLists.txt
@@ -137,17 +137,29 @@ NSCP_CREATE_TEST(
         ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
-# The licensing decode and the owner/DACL classification are platform neutral
-# (only the data acquisition is Windows), so these unit tests build everywhere.
+# The licensing decode, the owner/DACL classification and the command bodies of
+# the Windows-only posture checks are platform neutral (only the data
+# acquisition is Windows), so these unit tests build everywhere. The command
+# tests stand in for the platform sources, which is what lets the *matched* half
+# of these checks - thresholds, rendering, perf data - run off Windows at all.
 NSCP_CREATE_TEST(
     check_posture_test
     SOURCES
         check_activation_test.cpp
         check_file_security_test.cpp
         check_firewall_rules_test.cpp
+        check_posture_commands_test.cpp
         check_activation.cpp
         check_file_security.cpp
         check_firewall_rules.cpp
+        check_antivirus.cpp
+        check_bitlocker.cpp
+        check_defender.cpp
+        check_group_members.cpp
+        check_local_accounts.cpp
+        check_nla.cpp
+        check_secureboot.cpp
+        check_users.cpp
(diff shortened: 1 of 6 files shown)


Fantu:
@mickem from CI `CodeQL / Analyze (rust) (dynamic) Failing after 27s` seems missed to remove the trust part from the CI

PR #1481, opened 2026-09-04
feature: bring back .NET plugins, hosted through hostfxr on Windows and Linux

mickem:
The DotnetPlugins module had been switched off (module.cmake set BUILD_MODULE 0
on Windows) because it was a C++/CLI mixed-mode build: /clr forced a dynamic
CRT and asynchronous exceptions onto one module in an otherwise static build,
the managed API was a C++/CLI header compiled into NSCP.Core.dll, the protobuf
messages were compiled with a checked-in Google.Protobuf.dll through the .NET
Framework csc, and every one of those pieces broke the Windows link in a
different way over the years. None of it could ever build anywhere else.

Replace the whole approach with the runtime's own native hosting API:

* modules/DotnetPlugins is now plain C++ with no build-time dependency on .NET.
  When it loads it finds an installed runtime matching the process architecture
  (the `runtime path` setting, DOTNET_ROOT_<ARCH>/DOTNET_ROOT, the Windows
  registry, /etc/dotnet/install_location, the usual install folders, and the
  dotnet launcher on PATH last), picks the newest host/fxr/<version>/hostfxr
  library, starts the runtime through hostfxr_initialize_for_runtime_config and
  resolves the managed bridge with load_assembly_and_get_function_pointer. The
  hostfxr ABI comes from the runtime's own hostfxr.h / coreclr_delegates.h
  (MIT, vendored under modules/DotnetPlugins/hostfxr), so the calling
  conventions that matter on 32-bit Windows are the upstream ones. The runtime
  lives once per process and survives module reloads.
* libs/dotnet-plugin-api is a C# SDK project, NSCP.Core.csproj (net8.0,
  RollForward=LatestMajor), holding the plugin contract that used to be the
  C++/CLI NSCP.h, the NSCP.Helpers classes moved from libs/protobuf_net, the
  protobuf messages generated by protoc --csharp_out (Google.Protobuf now comes
  from NuGet), and NSCP.Core.Native.Bridge: [UnmanagedCallersOnly] entry points
  (Load/Start/Unload/Describe/Query/Submit/Exec/Message) with a C ABI, plus one
  core callback the other way (query/exec/submit/reload/settings/registry/log).
  Responses travel through a write callback so no memory ownership crosses the
  boundary. Each plugin gets its own collectible AssemblyLoadContext that
  resolves private dependencies from the plugin's deps.json but shares
  NSCP.Core and Google.Protobuf with the host so IPluginFactory is one type on
  both sides. All four handler kinds reach a plugin: queries and channel
  registrations are observed in the registry traffic so commands and passive
  results route back to the plugin that registered them, `exec` commands are
  routed by module name, and log entries are handed to plugins with a message
  handler.
* modules/CSharpSamplePlugin is an ordinary csproj referencing the built
  NSCP.Core.dll, exactly the way a third-party plugin is built, and implements
  all four handlers. It is built into <build>/modules/dotnet for the tests but
  is not installed or packaged: it is an example, not part of the product. Its
  response was never flushed (CodedOutputStream without Flush) - it now uses
  ToByteArray(). Helpers.registerKey never attached its request to the message.
* CMake: FindDotnet.cmake locates a dotnet SDK of version 8 or later
  (NSCP_DOTNET=OFF disables the managed build; an older SDK skips it);
  NSCP_ADD_DOTNET_PROJECT runs `dotnet build` into <build>/modules/dotnet with
  obj/bin redirected into the binary dir; WRAP_PROTO emits C# on every platform
  when the SDK is present. For distribution builds without network access the
  restore is steerable: NSCP_DOTNET_PACKAGE_SOURCE (a local folder with the
  .nupkg), NSCP_DOTNET_NO_RESTORE (packages already in $NUGET_PACKAGES) and
  NSCP_DOTNET_PROTOBUF_VERSION (when the host's protoc is newer than the
  default runtime pin). The legacy Find/Use{CSharp,Mono,DotNetFrameworkSdk}.cmake
  modules, libs/protobuf_net and the checked-in Google.Protobuf.dll are gone
  (REUSE.toml and the notices updated). The native module builds everywhere;
  only modules/dotnet needs the SDK.
* Packaging: the managed API installs as an explicit four-file list
  (NSCP.Core.dll, its deps.json and runtimeconfig.json, Google.Protobuf.dll)
  into modules/dotnet next to the native modules on Linux, and the DEB/RPM
  carry a Suggests on dotnet-runtime-8.0; the MSI's ".NET plugin support"
  feature is back, with the managed files gated on -dDotNet from CMake so a
  build without the SDK still produces an MSI. installer_feature_hint knows
  the module.
* Configuration moves to the module's own section: /settings/dotnet (plugin
  path, runtime path, factory class) and /settings/dotnet/plugins with
  <alias> = <assembly>|enabled|disabled, and a per-plugin `factory class`
  override under plugins/<alias>.

Tests: dotnet_plugins_test covers the version parsing/ordering, the hostfxr
locator on synthetic install trees, PE/ELF architecture detection,
DOTNET_ROOT/override precedence, the host singleton's failure paths and the
assembly name resolution (configured spelling first, then .dll; .exe and
extensionless assemblies load). tests/dotnetplugins.test.ts drives the real
binary over the client-query path: the C# sample answers check_dotnet, reads a
setting and logs through the core, round-trips an exec command and a passive
submission, sees log entries in its message handler, and a disabled plugin, a
missing assembly and a wrong factory class are reported without taking the
module down. CI installs dotnet-sdk-8.0 for the Debian/RPM builds and the
matching-architecture dotnet-runtime-8.0 for the Linux and Windows integration
tests; the build jobs stage the sample plugin next to the package in the
workflow artifact and the integration jobs copy it into the installed
modules/dotnet, since the packages themselves do not carry it.

Docs: generated reference page for the module (now under generic, it is no
longer Windows-only), an "Extending with .NET" walkthrough, the build
dependency table and an upgrade note.

Assisted-by: Claude Code:claude-fable-5-1
Claude-Session: https://claude.ai/code/session_01BzcSkCXZL4H5gkTXr62t6d
Signed-off-by: Michael Medin <michael@medin.name>


mickem:
@Fantu could you check if this builds in your debian setup.

Fantu:
dotnet is not present in the Debian repository. I won't add it to Debian (so I can have the dedicated module in nscp). If I remember correctly, I've seen people wanting to package it several times, but it was a huge and difficult task.
Furthermore, even if it were in the repo, I think I'd put the module in a separate package at most, as recommended (maybe not even), as it adds over 100MB of runtime (just the most minimal part) and I assume it would be rarely used.
It would be interesting to see if it could be made compatible with Mono.
I'll have Claude look deep into it because, after a quick look, I can think of some potential problems that need to be verified (what I mean is not related to possible mono compatibility).

Fantu:
Here's the result of Claude's analysis, and his response is below. Be warned, I haven't read and verified everything thoroughly (I have a lot of other things to do and don't want to spend hours on this part). I'm reporting it in full, however, and I'll let you know that it's not verified.
From what I've quickly read about Mono, I suppose it's best to exclude it.
I think that in Debian I'll disable this part in a future packages, as mentioned above.
To prevent various other distros from doing the same (if they package nscp in the future), I think the Google.Protobuf issue should be looked into.

--------

I built and looked into this on Debian. Summary below: build result, the Mono question, then what I think is worth changing.

## Build on Debian sid

Full build in a sid chroot (gcc 16.2, plus the three GCC 16 fixes from my `fix-gcc16-build` branch, which are not in this PR): builds clean, 86/87 tests pass (the one failure is `PathManagerTest.CaPathExpandsToBundleFile`, unrelated: the chroot has no `ca-certificates` and I pass `-DCONFIG_CA_PATH`).

Without the SDK everything degrades the way it should: configure reports `! dotnet SDK not found`, only the native module is built (496 KB stripped) and installed, no `modules/dotnet`. With a plugin configured and no runtime present the module reports

```
E dotnet No x64 .NET runtime found (looked for host/fxr/<version>/libhostfxr.so under: ...).
```

and the agent keeps running. That part is nicely done.

## Mono

I checked whether the managed side could run on Mono, since that is what Debian actually ships. It cannot, without a second implementation. Tested against mono 6.14.1 in sid:

- `[UnmanagedCallersOnly]` — `error CS0246: type or namespace could not be found`
- `System.Runtime.Loader.AssemblyLoadContext` — `error CS0234: 'Loader' does not exist in namespace 'System.Runtime'`
- `delegate* unmanaged[Cdecl]<>`, `Marshal.PtrToStringUTF8`, `ReadOnlySpan<byte>` — same story, .NET Core 3.0/5+ only
- Mono has no hostfxr at all; it only exposes the embedding API (`mono_jit_init` & co.)
- SDK-style csproj cannot be built: Debian has no `msbuild` and no `nuget`, only `xbuild` (Engine 14.0), with an empty `xbuild-frameworks` and no `netstandard.dll`
- there is no CIL Google.Protobuf package in Debian

So Mono support would mean a second native host on the embedding API, a second managed API target without `UnmanagedCallersOnly`/ALC/`deps.json`, and a Google.Protobuf for CIL that would have to be packaged first. For context on where that road leads: the `mono` source package builds 168 binary packages in trixie and 5 in sid. I would not invest there.

## Things I would change

**1. The managed build restores from nuget.org.** `NSCP.Core.csproj` has `<PackageReference Include="Google.Protobuf" Version="3.36.1" />` and `NSCP_ADD_DOTNET_PROJECT` runs `dotnet build` without `--no-restore`, so the build needs network access (and a writable `$HOME` for `~/.nuget`). That is not allowed in Debian, Fedora or openSUSE build environments, which means no distribution package will ever be able to ship the .NET plugins — only your own binaries can. `build.md` documents `-DNSCP_DOTNET=OFF`, which is the right escape hatch, but it would be worth also supporting an offline restore (honouring `NUGET_PACKAGES` / `-p:RestorePackagesPath=` / `RestoreSources=` pointing at a local folder, and `--no-restore` when the packages are already there). Small change, and it keeps the door open for packagers.

**2. The managed install is a glob, not a file list.** `libs/dotnet-plugin-api/CMakeLists.txt` installs `${NSCP_DOTNET_OUTPUT_DIR}/` with `FILES_MATCHING PATTERN "*.dll" PATTERN "*.json"`. That directory is also where the C# sample and anything else built into the tree lands, so the package contents depend on what happened to be in the build directory, and `NSCP.Plugin.CSharpSample.dll` + `.deps.json` end up in production packages (the MSI lists them explicitly too). I would list the files like `Product.wxs` does, and keep the sample out of the packages or behind its own feature.

**3. No runtime dependency declared on the packages.** CI now installs `dotnet-sdk-8.0` for the Debian and RPM builds, so the official packages will contain `NSCP.Core.dll`/`Google.Protobuf.dll` while nothing suggests `dotnet-runtime-8.0` to the user. Worth at least a Recommends/Suggests, or a separate subpackage: the minimal runtime closure is ~104 MB installed (dotnet-runtime-8.0 66.6 MB, libicu 35.1 MB, hostfxr/host/lttng/numa the rest) on Ubuntu 24.04.

**4. `Google.Protobuf` is pinned while the C# is generated by the system protoc.** `WRAP_PROTO --csharp_out` uses whatever protoc the build host has, but the runtime is pinned to 3.36.1. On sid protoc is 3.21.12 and matches your comment; on distributions with a much newer protobuf (Fedora, openSUSE) the generated code may want a newer runtime than the pin. Worth checking, or making the version overridable.

**5. Minor:** the PR carries one unrelated commit (`test(NRDPClient): drain the request before answering 403`).

## What already looks right

`reuse lint` still passes on the branch (2357/2357 files covered; the only complaint is the pre-existing unused OFL-1.1/Zlib texts): the vendored hostfxr headers are annotated MIT in `REUSE.toml` and listed in `THIRD-PARTY-NOTICES.md`, and `tools/license_headers.py` excludes them. Dropping the checked-in `Google.Protobuf.dll` binary is a real improvement for distributions. License texts and notices are already shipped in the Linux packages by `files/CMakeLists.txt`, so redistributing Google.Protobuf is covered — the only nit is that the notices row describes it as "NuGet package restored at build time" when it is also shipped in `modules/dotnet`. The integration tests degrade instead of failing when the runtime or the managed API is missing, and `dotnet_plugins_test` passes with no .NET installed.

mickem:
Thanks @Fantu, that was a useful review. Pushed in 1e86be9, on top of a merge with main:

1. **Offline restore**: `NSCP_ADD_DOTNET_PROJECT` now forwards `-DNSCP_DOTNET_PACKAGE_SOURCE=<dir|url>` as `RestoreSources` and `-DNSCP_DOTNET_NO_RESTORE=ON` as `--no-restore`; `NUGET_PACKAGES` and a `NuGet.Config` pass through as before. Verified with an empty package cache and a folder holding only `google.protobuf.3.36.1.nupkg`.
2. **Install list**: the API installs exactly `NSCP.Core.dll`, `.deps.json`, `.runtimeconfig.json` and `Google.Protobuf.dll`, no glob. The sample installs its own two files behind `NSCP_DOTNET_INSTALL_SAMPLE` (default ON, since the Linux integration tests drive the installed sample and the packages already ship the sample scripts; `-DNSCP_DOTNET_INSTALL_SAMPLE=OFF` keeps it out). In the MSI it moved into its own component under the Scripts feature.
3. **Runtime**: DEB and RPM carry `Suggests: dotnet-runtime-8.0` when the managed API is included. A subpackage split is left to the maintainer.
4. **Protobuf pin**: `-DNSCP_DOTNET_PROTOBUF_VERSION=` overrides the `Google.Protobuf` version in both csproj files (default still 3.36.1).
5. The NRDP test commit stays in this PR: it fixes a Windows-only flake in main's test that kept this PR's CI red.

The notices row now says the runtime is also shipped in `modules/dotnet`. Mono is out of scope, as you concluded.

---

mickem:
Update on point 2: as of 7c1c150 the C# sample is no longer shipped at all. It is still built into `<build>/modules/dotnet` for the integration tests, but neither the DEB/RPM/ZIP nor the MSI carry it (the `NSCP_DOTNET_INSTALL_SAMPLE` option is gone again). The packages contain exactly `NSCP.Core.dll`, `NSCP.Core.deps.json`, `NSCP.Core.runtimeconfig.json` and `Google.Protobuf.dll` under `modules/dotnet`; CI hands the sample to the integration jobs through the workflow artifact instead.

---

Issue #1479, opened 2026-09-04
FTBFS with GCC 16: the C++ standard is never declared, so a compiler default change breaks the build


Fantu:
`build.md` lists a "C++17 toolchain" as the requirement, but the build never tells CMake: there is no `CMAKE_CXX_STANDARD` (nor `target_compile_features`) anywhere, so the language level is whatever the compiler happens to default to.

Debian unstable now ships GCC 16 (`g++ 16.2.0`), whose default is `-std=gnu++20`, and 0.18.1 does not compile there:

```
modules/CauseCrashes/module.cpp:183:80: error: no matching function for call to
  'nscapi::plugin_instance_data<CauseCrashesModule>::plugin_instance_data(<brace-enclosed initializer list>)'
  183 |         static nscapi::plugin_instance_data<plugin_impl_class> plugin_instance{};
      |                                                                                ^
```

`plugin_instance_data` (`include/nscapi/nscapi_plugin_wrapper.hpp:27`) declares its copy constructor `= delete`. In C++17 the class is still an aggregate, so the `plugin_instance{}` line that `build/python/create_plugin_module.py` generates for **every** module is aggregate-initialised; in C++20 a user-declared constructor disqualifies the aggregate, and the deleted copy constructor left no default one to call. CauseCrashes is just the first module compiled.

This is not a 0.18.1 regression — the `= delete` dates back to `121587ff` (February) — and CI cannot see it: the Linux jobs build on Ubuntu 24.04 and Rocky, where the compiler still defaults to C++17.

Two things would help:

1. **Declare the standard** (`set(CMAKE_CXX_STANDARD 17)` + `CXX_STANDARD_REQUIRED`), so the language cannot move under the project every time a distribution bumps its compiler.
2. **Decide what is supported.** Is C++17 the floor, and is building as C++20 meant to work? Newer toolchains keep arriving — Debian unstable already, other distributions will follow — so it would help to know which compilers you want green, and perhaps to have one CI job on a rolling distribution catching this class of breakage early.

The minimal fix for this particular error is a defaulted constructor:

```c++
plugin_instance_data() = default;
```

Fantu:
Working past this one turned up a second, independent failure with the same compiler. Posting it here rather than in a new issue since it is the same "does not build with GCC 16" story — happy to split it out if you prefer to track it separately.

Forcing `-std=gnu++17` gets the build to 93%, where both standalone clients fail to link:

```
undefined reference to `nscapi::plugin_singleton'
[...]

PR #1483, opened 2026-09-04
Fix the build with GCC 16

Fantu:
Two independent build failures with GCC 16 (Debian unstable, `g++ 16.2.0`), reported in #1479. Neither is a regression: both reproduce on 0.14.1, and CI cannot see them because the Linux jobs run on Ubuntu 24.04 and Rocky.

- **The plugin instance no longer aggregate-initialises.** GCC 16 defaults to `-std=gnu++20`, where `plugin_instance_data` is not an aggregate any more — its copy constructor is `= delete`, which also suppressed the implicit default one — so the `plugin_instance{}` line generated for *every* module has no constructor to call. A defaulted constructor makes it valid under both standards.
- **`nscapi::plugin_singleton` is missing from the standalone clients.** `check_nscp` and `check_nscp_nrpe` include a module client handler whose inline members log through `GET_CORE()`. They never call them, but GCC 16 emits them anyway (speculative devirtualization) and the link fails. Defined in both, the way the unit-test binaries already do.

Verified: with both applied, a full Debian package build on unstable succeeds with GCC 16 in its default C++20 mode, 70/70 tests passing.

This does not settle the wider question in #1479 — the project still never declares a C++ standard, so the language level follows whatever the compiler defaults to. Happy to add `set(CMAKE_CXX_STANDARD ...)` here if you tell me which one you want.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ project(
     C
 )
 
+# Build against a known language level instead of whatever the compiler happens
+# to default to: that default moves (GCC 16 switched to C++20).
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 include(CTest)
 if(COMMAND CMAKE_POLICY)
     cmake_policy(SET CMP0011 NEW)
--- a/clients/nrpe/check_nrpe.cpp
+++ b/clients/nrpe/check_nrpe.cpp
@@ -17,6 +17,10 @@
 #include "../../modules/NRPEClient/nrpe_client.hpp"
 #include "../../modules/NRPEClient/nrpe_handler.hpp"
 
+// Never called - this client uses its own handler - but the module handler
+// included above references it, and an optimizing compiler emits it.
+nscapi::helper_singleton *nscapi::plugin_singleton = new nscapi::helper_singleton();
+
 std::string gLog = "";
 
 int main(int argc, char *argv[]) {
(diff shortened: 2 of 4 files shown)


mickem:
Thanks.

After reviewing this I understand the problem and think it makes sense to  `set(CMAKE_CXX_STANDARD ...)` set to 17 while we are at it? 
Do you want to add that to this PR?

mickem (approved):
Thanks

Fantu:
Done — added as a third commit.

`CMAKE_CXX_STANDARD 17` with `CMAKE_CXX_STANDARD_REQUIRED ON`, so a compiler that cannot do C++17 fails at configure time instead of quietly falling back. I left `CMAKE_CXX_EXTENSIONS` at its default, so GCC and Clang get `-std=gnu++17` — exactly what they picked on their own before GCC 16 moved.

Rebuilt on Debian unstable with GCC 16.2 (full package build, `-O2` plus the distro hardening flags): the tree compiles at `-std=gnu++17`, both standalone clients link, and ctest is 86/86.

Two things worth your eye, since I can only build on Linux:

- **MSVC.** Its default is `/std:c++14`, so this does raise the level on Windows; v141 and v141_xp both accept `/std:c++17`, but the legacy job is the one that would tell us if something in the XP toolset's SDK disagrees.
- **`installer_lib`** is configured as a separate top-level project, so it is untouched and still builds at the compiler default. Say the word if you want the same two lines there.

Unrelated, noticed while running the suite: `PathManagerTest.CaPathExpandsToBundleFile` asserts that the compiled-in `ca-path` exists *on the build machine*, so it fails in a build chroot that has no `ca-certificates` installed. Worth relaxing to a path check, or skipping when the file is absent?

Fantu:
@mickem can you please consider this PR for next release?

mickem:
Absolutely... Will look it over and get it in later today...

PR #1484, opened 2026-09-04
docs: drop leftover Sphinx roles from the plugin API page

Fantu:
`docs/docs/extending/plugin-api.md` still carries eight `:py:class:` roles from the reStructuredText days. They used to render as literal text, but since `pymdownx.emoji` was enabled `:py:` is read as an emoji shortcode, so every message-type reference on the page now shows a Paraguay flag pulled from the twemoji CDN instead of the class name.

Dropping the role prefix leaves the backticked name, which is what the rest of the page already uses.

It is also the only remote asset left in the generated site, which matters to distributions shipping the manual offline — Debian's lintian reports it as a privacy breach. The `:fontawesome-*:` icons in the reference index are unaffected: those are bundled with mkdocs-material and inlined as SVG.

### DIFF
--- a/docs/docs/extending/plugin-api.md
+++ b/docs/docs/extending/plugin-api.md
@@ -7,7 +7,7 @@
 
 `Plugin.QueryRequestMessage` Query request
 Used for querying the client this is the "normal" check_nrpe message request.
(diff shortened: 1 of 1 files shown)


mickem (approved):

mickem:
thanks

PR #1485, opened 2026-09-04
fix(CheckSystemUnix): take the dpkg install date from dpkg-query

Fantu:
`check_installed_software` dated every installed package by `stat()`ing `/var/lib/dpkg/info/<name>[:<arch>].list`. That is dpkg's internal database rather than an interface: which file belongs to a package depends on its `Multi-Arch` field and on the database format version, so the code had to guess, trying `<name>:<arch>.list` and falling back to `<name>.list`.

`dpkg-query` exposes the same timestamp as `${db-fsys:Last-Modified}` (since dpkg 1.19.3), resolved by dpkg itself, so this asks for it in the query already being made and drops the stat pass. It returns exactly the mtime the old code read. This lines the dpkg path up with the rpm one, which has always taken `%{INSTALLTIME}` from rpm, and as a consequence silences lintian's `uses-dpkg-database-directly` warning on the built module.

Behaviour changes:

- An older `dpkg-query` leaves the column empty and still exits 0, so `install_date` stays unset there (unit test and upgrade note).
- dpkg-query aborts on a `stat()` error other than `ENOENT`, so an unreadable dpkg database now makes the check UNKNOWN instead of leaving one package undated (described in the commit message).

Also updates the check's documentation and adds an integration test that every package gets a date on dpkg and rpm hosts.

### DIFF
--- a/modules/CheckSystemUnix/check_installed_software.cpp
+++ b/modules/CheckSystemUnix/check_installed_software.cpp
@@ -3,7 +3,6 @@
 
 #include "check_installed_software.h"
 
-#include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
@@ -68,12 +67,6 @@ command_result run_command(const std::string &cmd) {
 
 bool binary_exists(const std::string &path) { return access(path.c_str(), X_OK) == 0; }
 
-long long file_mtime(const std::string &path) {
-  struct stat st{};
-  if (stat(path.c_str(), &st) != 0) return 0;
-  return static_cast<long long>(st.st_mtime);
-}
-
 // "Name <email>" → "Name" (the email part adds noise to publisher filters).
 std::string strip_email(const std::string &maintainer) {
   const std::size_t angle = maintainer.find(" <");
@@ -112,7 +105,7 @@ package_manager detect_manager() {
 }
 
 // Parse dpkg-query -W output with the format
-//   ${Package}\t${Version}\t${Architecture}\t${Maintainer}\t${Installed-Size}\t${Status}
+//   ${Package}\t${Version}\t${Architecture}\t${Maintainer}\t${Installed-Size}\t${Status}\t${db-fsys:Last-Modified}
 // Status is three words ("<desired> <error> <state>"). Only a state of exactly
 // "installed" counts: the states that merely end in it ("not-installed",
 // "half-installed") are removed or broken packages, not installed ones.
@@ -142,6 +135,17 @@ std::vector<software_entry> parse_dpkg_output(const std::string &output) {
     } catch (...) {
       e.size_bytes = 0;
     }
+    // The install date is the last field. dpkg-query grew db-fsys:Last-Modified
+    // in 1.19.3; an older one leaves the column empty, which reads as "unknown"
+    // just like a manager that records no date at all.
+    if (parts.size() >= 7) {
+      try {
+        e.install_date_epoch = std::stoll(boost::trim_copy(parts[6]));
+      } catch (...) {
+        e.install_date_epoch = 0;
+      }
+    }
+    e.install_date_str = format_epoch_date(e.install_date_epoch);
     if (e.name.empty()) continue;
     out.push_back(e);
   }
@@ -204,23 +208,12 @@ std::vector<software_entry> parse_pacman_output(const std::string &output) {
   return out;
 }
 
-void apply_dpkg_install_dates(std::vector<software_entry> &entries, const mtime_fn &mtime) {
-  for (software_entry &e : entries) {
-    if (e.manager != "dpkg" || e.install_date_epoch > 0) continue;
-    // Multi-arch packages use <name>:<arch>.list; older/same-arch ones plain <name>.list.
-    long long t = 0;
-    if (!e.architecture.empty()) t = mtime("/var/lib/dpkg/info/" + e.name + ":" + e.architecture + ".list");
-    if (t == 0) t = mtime("/var/lib/dpkg/info/" + e.name + ".list");
-    e.install_date_epoch = t;
-    e.install_date_str = format_epoch_date(t);
-  }
-}
-
 fetch_result fetch_installed(const package_manager &manager, const exec_fn &exec) {
   fetch_result result;
   if (manager.name == "dpkg") {
     const command_result r =
-        exec(manager.binary + " -W -f='${Package}\\t${Version}\\t${Architecture}\\t${Maintainer}\\t${Installed-Size}\\t${Status}\\n' 2>/dev/null");
+        exec(manager.binary +
+             " -W -f='${Package}\\t${Version}\\t${Architecture}\\t${Maintainer}\\t${Installed-Size}\\t${Status}\\t${db-fsys:Last-Modified}\\n' 2>/dev/null");
     result.ok = r.ok;
     if (r.ok) result.entries = parse_dpkg_output(r.output);
     return result;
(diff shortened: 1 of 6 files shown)


modules/CheckSystemUnix/check_installed_software.cpp
```diff
@@ -142,6 +135,17 @@ std::vector<software_entry> parse_dpkg_output(const std::string &output) {
...
+    // The install date is the last field. dpkg-query grew db-fsys:Last-Modified
+    // in 1.19.3; an older one leaves the column empty, which reads as "unknown"
+    // just like a manager that records no date at all.
+    if (parts.size() >= 7) {
+      try {
+        e.install_date_epoch = std::stoll(boost::trim_copy(parts[6]));
+      } catch (...) {
+        e.install_date_epoch = 0;
+      }
+      e.install_date_str = format_epoch_date(e.install_date_epoch);
```
mickem, line 147: The `>= 7` guard is correct — I checked what an old dpkg-query actually emits. In 1.18.25/1.19.0.5 `pkg_format_show()` looks the name up in `fieldinfos`, then `virtinfos`, then the arbitrary fields; on a miss `ok` stays false and the field node contributes nothing, with no warning and exit 0. The literal `\t` between the fields is a separate `PKG_FORMAT_STRING` node and is still printed, so an old dpkg gives 7 fields with an empty last one, not 6. Reproduced the same shape on 1.22.6 with a bogus `${db-fsys:Bogus-Field}`: empty column, empty stderr, exit 0.

Nit: `install_date_str` is assigned inside the branch, so a 6-field line leaves it at its default. That is the correct value, but it reads as if the branch owned the invariant. `parse_rpm_output` sets it unconditionally right after the `try`/`catch` — worth matching that shape so the two parsers stay symmetric:

```cpp
if (parts.size() >= 7) {
  try {
    e.install_date_epoch = std::stoll(boost::trim_copy(parts[6]));
  } catch (...) {
    e.install_date_epoch = 0;
  }
}
e.install_date_str = format_epoch_date(e.install_date_epoch);
```

---

modules/CheckSystemUnix/check_installed_software.cpp
```diff
@@ -204,23 +208,12 @@ std::vector<software_entry> parse_pacman_output(const std::string &output) {
...
-  }
-}
-
 fetch_result fetch_installed(const package_manager &manager, const exec_fn &exec) {
   fetch_result result;
   if (manager.name == "dpkg") {
     const command_result r =
-        exec(manager.binary + " -W -f='${Package}\\t${Version}\\t${Architecture}\\t${Maintainer}\\t${Installed-Size}\\t${Status}\\n' 2>/dev/null");
+        exec(manager.binary +
+             " -W -f='${Package}\\t${Version}\\t${Architecture}\\t${Maintainer}\\t${Installed-Size}\\t${Status}\\t${db-fsys:Last-Modified}\\n' 2>/dev/null");
```
mickem, line 216: One behavioural difference worth recording somewhere, because it is invisible from this line: `virt_fsys_last_modified()` returns an empty field on `ENOENT`, but calls `ohshite()` on any other `stat()` failure, which makes dpkg-query exit non-zero. `fetch_installed` then reports `r.ok == false` and the check turns into UNKNOWN "Failed to query installed software from dpkg". Before this change a stat failure just left that one package's date at 0 and the rest of the inventory still came back.

It needs a non-`ENOENT` error on `/var/lib/dpkg/info`, so it is narrow, and failing loudly on a broken package database is arguably the right call given the check already refuses to report a clean OK on a failed query. But `2>/dev/null` means the operator never sees dpkg's reason for it. Either mention the trade-off in the commit message, or keep dpkg's stderr and fold it into the error message.

---

mickem (commented):
Good change, and I think the justification is stronger than the one the description leads with. I verified the claims rather than taking them on trust:

- **The field is the same value.** On a live 1.22.6 host `dpkg-query -W -f='${db-fsys:Last-Modified}'` matches `stat -c %Y` on the `.list` file exactly (checked bash, libc6, adduser).
- **1.19.3 is the right floor.** `db-fsys:Last-Modified` is in `virtinfos[]` in 1.19.3's `lib/dpkg/pkg-format.c` and absent from 1.19.0.5.
- **The old-dpkg fallback holds**, and gives 7 fields with an empty last one rather than 6 — details inline.
- **dpkg's implementation is a single `stat()`** on the path from `pkg_infodb_get_file(pkg, pkgbin, LISTFILE)`, i.e. dpkg resolves the multi-arch name itself with `pnaw_nonambig`. That is the real win here: the `<name>:<arch>.list` fallback was a guess about dpkg's internal layout, and this replaces it with dpkg's own answer.

Four things I would like addressed before merge.

### 1. The docs still describe the old mechanism

`docs/samples/CheckSystemUnix_check_installed_software_desc.md` (the **Caveats** paragraph) still says:

> dpkg does not record them, so they are approximated from the mtime of the package's `/var/lib/dpkg/info/<name>[:<arch>].list` file

That text is merged into the generated reference page, so it ships as documentation of a code path that no longer exists. It should say the date comes from `dpkg-query`'s `db-fsys:Last-Modified`. The "rewritten on upgrade — treat as last installed/upgraded" caveat stays true, since it is still that same mtime, and it is worth adding that a dpkg older than 1.19.3 leaves the field unset.

### 2. This needs an upgrade note

Debian 9 ships dpkg 1.18.25 and Ubuntu 18.04 ships 1.19.0.5, both below the floor. On those hosts `install_date` silently goes from populated to empty, so a policy like `warning=install_date > -7d` stops matching and the check just goes quiet — the failure mode is a check that looks healthy while no longer testing anything.

That is the "behaviour change that needs operator action" case in `CLAUDE.md`, so it wants a new `docs/upgrades/0.18.2/<slug>.md` (or whichever version this lands in) with `modules: CheckSystemUnix` and `action: conditional`: affected only if you run dpkg older than 1.19.3 *and* filter on `install_date`.

### 3. The performance framing overclaims

The syscalls do not go away, they move: dpkg now does one `stat()` per package where the module did one or two. Measured on a 686-package host, adding the field costs dpkg-query about 1.4 ms per run (41 ms → 48 ms over five runs), against up to 2N stats dropped from the module — so a small net win, not "a few thousand extra syscalls" saved.

I would lead with the interface argument instead, which is the one that actually justifies the change: the `.list` path was reverse-engineered from dpkg's internal database, dpkg answers it authoritatively, and lintian's `uses-dpkg-database-directly` goes quiet as a consequence rather than as the point.

### 4. No integration test covers the date

The unit test is right, but nothing under `tests/` would have gone red if `install_date` silently stopped being populated — `checksystem-commands.test.ts` exercises the command without asserting on the date. This change and its pre-1.19.3 fallback are exactly the class of breakage the integration tests exist for. A Linux-guarded case that renders `install_date` through a `top-syntax` and asserts it is non-empty is cheap and deterministic on any dpkg runner (every installed package has a `.list` mtime).

---

Fantu:
Thanks for the thorough review. Rebased on main and updated:

1. Docs: the Unix sample was folded into `CheckSystem_check_installed_software_desc.md` in the meantime, so the Linux caveat there now describes `db-fsys:Last-Modified` and the unset date on dpkg < 1.19.3.
2. Upgrade note: added under `docs/upgrades/next/` (per the current README, rather than a version directory), `CheckSystemUnix`, `action: conditional`.
3. Agreed, the syscall argument was weak. The description and commit message now lead with the interface one.
4. Added a Linux case to `checksystem-commands.test.ts` that goes critical on any dpkg/rpm entry with an empty `install_date_s`.

Inline: `install_date_str` is now set after the `if`, as in the rpm parser. For the non-`ENOENT` case I went with the commit message (and the upgrade note): capturing dpkg's stderr apart from the parsed output did not seem worth it for a broken database.

mickem:
Thank you

PR #1492, opened 2026-09-05
Guard client credentials and secure script uploads

mickem:
## Summary

This PR implements three security hardening measures addressing credential leakage, predictable file staging, and directory junction attacks:

1. **Client credentials stay with their target** — Configured targets carrying passwords or tokens now refuse `host=`, `port=`, and `address=` arguments from requests, preventing credential redirection to caller-chosen hosts.

2. **Private script-upload staging** — REST script uploads are now staged in randomly named, exclusively created files instead of predictable paths in the shared temp directory.

3. **Junction-proof shared folder lockdown** — The modern-layout `%ProgramData%\NSClient++` protection now detects and refuses reparse points (junctions/symlinks), securing the directory entry itself rather than following links to their targets.

### Client Credential Guard
- Added `is_sensitive_key()` to identify password/token credentials
- Extended `destination_container` with `configured_target`, `configured_credentials`, and `allow_host_override` fields to track credential-bearing targets
- Implemented `configuration::check_host_override()` to validate that credentialed targets refuse destination overrides
- Applied the guard to both query (`do_query`) and execute (`do_exec`) paths
- Targets can opt in to the old behavior with `allow host override = true`
- Added 13 comprehensive tests covering all scenarios: credentialed targets, empty credentials, certificates, target selection, and REST-style arguments

### Script Upload Staging
- Created `upload_staging` namespace with `create_exclusive()` and `stage()` functions
- `create_exclusive()` uses `O_EXCL`/`CREATE_NEW` and `O_NOFOLLOW` to prevent symlink attacks and file reuse
- Files are created with owner-only permissions (0600 on POSIX)
- All writes are checked; staging failures are reported rather than silently ignored
- Updated `scripts_controller.cpp` to use the new staging mechanism
- Added 8 unit tests covering random naming, binary content, empty files, permissions, and symlink refusal

### Directory Junction Protection
- Added `open_plain_directory()` helper to open directory entries without following reparse points
- Added `is_reparse_point()` to detect junctions/symlinks before attempting protection
- Modified `protect_directory()` to use handle-based `SetSecurityInfo()` instead of path-based `SetNamedSecurityInfo()`
- Updated `NSClient++.cpp` to check for and reject reparse points at the shared folder path
- The check is fatal: a planted junction prevents service startup rather than being "secured through"

## Documentation

- Updated `docs/docs/security/notices.md` with detailed security notice explaining each issue, impact, and mitigation
- Updated `docs/docs/setup/upgrading.md` with migration guidance for the credential guard
- Updated `docs/docs/api/rest/queries.md` with a note about client-module credential behavior

## Testing

- 13 new tests in `command_line_parser_test.cpp` for the credential guard
- 8 new tests in `upload_staging_test.cpp` for file staging
- Tests cover both success and failure paths, edge cases, and security properties

https://claude.ai/code/session_01NLdzs5rC7975A21teDm6sB


Fantu:
@mickem I asked Claude to do a PR analysis, and he did a thorough review and many more tests than I expected.
He identified several things. I don't have time to read and check everything, so I'll let you know. I'm posting the message he generated in case you want to look it up and check it:

--------------
Read through this while checking what it means for the Debian package (nothing, as it happens: WEBServer is not built there and the ACL work is Windows-only, so what ships is the credential guard). I also dug into the red CI, since all four failing job families look like they belong to this branch.

## What the failing CI is saying

**The integration tests (ubuntu 24.04 and rocky 10, both architectures): 97/548, all in the NSCA suite, all the guard.** Every failure is `The target '…' carries credentials, so its destination cannot be changed from the command line (--address)`. The caller is `scripts/python/test_nsca.py`'s `submit_via_exec()`, which passes `--target <t>` together with `--address`, `--encryption`, `--password` and `--payload-length` — that is, it names a destination *and brings its own credential for it*. The configured target's password would never have travelled anywhere, yet the call is refused.

That looks like the guard's rule is one step too coarse: what matters is not whether the target carries a credential, but whether the *configured* credential would be the one going to the caller-named host. If the command line also sets the sensitive keys for that target, there is nothing left to protect and the call could proceed. Otherwise every caller who does "one target, address and secret per call" — which the project's own suite does — has to set `allow host override = true`, i.e. switch the guard off entirely.

(Side note from the same log: for target `test_rp` the message names `'default'`. `test_rp` is not a configured target, `get_target()` falls back to `default`, and the refusal then cites a target the caller never asked for. Confusing in the message, though the guard itself is behaving as written.)

**sanitizer-tests: the overflow is in the new test, not in the code under test.** `upload_staging_test.cpp:68` builds

```c++
std::string content("\x00\x01\r\n\xff\xfe binary \x00", 17);
```

but that literal is 15 bytes, so 16 with its terminator — ASan reports a 17-byte read ending 0 bytes past a 16-byte global, which matches exactly. `15`, or `sizeof(lit) - 1`, fixes it; `stage()` itself is fine.

**The Windows builds (x64, x86, x86-static): `AclTest.ResettingMakesARenamedTreeInheritTheDestination`.** `is_protected()` is called there on a *file* inside the renamed tree, and the failure message is `the path is not a directory` — the new `open_plain_directory()` rejects anything without `FILE_ATTRIBUTE_DIRECTORY`, so `inspect_protection()` no longer answers for files, where the old path-based `GetNamedSecurityInfoW` did. The reparse-point check is the part that matters for inspection; requiring a directory as well narrows a function that had other callers.

## `protect_directory`: the combined open may fail in the case it exists for

Separate from the test above. The new code opens one handle with `READ_CONTROL | WRITE_DAC | WRITE_OWNER` and does both `SetSecurityInfo` calls on it. With `FILE_FLAG_BACKUP_SEMANTICS`, `SeRestorePrivilege` (which you enable) covers `WRITE_DAC` and `WRITE_OWNER`, but `READ_CONTROL` is covered by `SeBackupPrivilege`, which is not enabled. Against a folder someone else owns with a DACL that denies you — the scenario this function exists for — the open then fails and `protect_directory` returns false without repairing anything, where the path-based version would have taken ownership and gone on to fix the DACL. Neither `SetSecurityInfo` call needs `READ_CONTROL`, so dropping it from the requested access is probably the smallest fix.

## The remedy the guard suggests does not work on the query path

The refusal message and the upgrade note both tell the operator to "configure the other host as its own target and select it with `target=`". But `$target.id$` is only ever consumed in `i_do_exec`; `i_do_query` never applies it.

Checked against 0.18.1, so this predates the PR — two NSCA targets at different addresses, `default` and `foo`:

```
--query submit_nsca target=foo    → Failed to connect to: 10.255.255.2:5667   (default; foo ignored)
--query submit_nsca --target foo  → Failed to connect to: 10.255.255.2:5667   (same)
--exec  submit_nsca target=foo    → Failed to connect to: 10.255.255.1:5667   (foo, as expected)
```

The client modules route `check_*`/`submit_*` through `do_query` when they are invoked as a query, which is what a REST or NRPE caller gets. So on that path the suggested alternative silently does nothing, and the only way out left is turning the guard off. Fixing the underlying gap is its own change, but the message probably should not recommend something that does not work where the refusal is most likely to be met.

`selecting_another_configured_target_is_not_an_override` does not catch this because it selects the target through the header (`query_request(..., "backup")`), not through the `target=` argument.

## Does the guard mean to cover the header too?

`do_query` and `do_exec` apply `d.apply(t, request.header())` before anything is parsed, and `apply_host()` feeds `address` and every metadata key through `set_string_data()` — including `host` and `port`. `check_host_override()` looks only at the `variables_map`, so a destination moved that way is not refused; and when `use_header` is set the arguments are never parsed at all (`// TODO: Parse header here`), so the guard does not run.

I could not find a remote caller that reaches it — the header hosts are populated by internal paths (the submit-from-query chain, the Lua/Python `script_nscp` helpers) — so this may be out of scope on purpose. Worth saying so explicitly, since the comment in the header states the invariant more broadly than the code enforces it.

## The module list on the upgrade note is wider than the change

`modules:` lists NRPEClient, GraphiteClient, SyslogClient and CheckMKClient, none of which have a `password` or `token` option on their targets, so the guard can never trigger for them — with the new per-module filter, their users get a "check your setup" entry that does not apply. The targets that do carry credentials are NSCA, NSCA-ng, NRDP (`password`/`token`/`key`), NSCP, Icinga and SMTP.

## Smaller

- `allow host override = yes` parses as false: `to_bool()` accepts only `true`/`1`/`True`. Pre-existing, but it now decides whether a protection is on.

Things I checked that are fine: `boost::filesystem::unique_path` is still present and not deprecated in Boost 1.90, so the staging code is safe on current distributions; and the docs side validates — `docs/hooks/notes.py --check` passes on this branch and the long anchor the three notes link to does match the generated heading in `security/notices.md`.

mickem:
Thanks @Fantu — that review was accurate on every point, and the three red job families were all defects in this branch rather than anything pre-existing. All of it is addressed in 4427b93 and 8f6a567, and the run on 8f6a567 is green (29/29, including sanitizer-tests and the Windows integration tests, which had been skipped until the Windows build started passing again).

**The guard was too coarse, as you diagnosed.** It refused on "the target carries a credential" rather than "the credential that would travel is the configured one". It now decides on two facts: the destination differs from the one its target configured, *and* at least one credential still in the container is the target's rather than the request's. Any credential set by the request clears the mark for that key, so `--target x --address y --password z` — what `test_nsca.py` does — proceeds, because nothing configured is at stake. That was the whole 97/548.

**Deciding on the resolved address rather than on the `variables_map` also answers your header question.** A destination moved through `apply_host()` is now covered, since the comparison is against the address the target configured however the request changed it. It also stops refusing an override that names the address the target already had.

**`target=` on the query path** — you were right that this predates the PR and that recommending it was useless where the refusal is most likely. `i_do_query` now applies `$target.id$` like `i_do_exec` always did, re-parsing the command line afterwards so an explicit option still beats the selected target. Confirmed by hand: `--query submit_nsca target=foo` and `--target foo` both reach foo now, where both previously used `default`.

**The ASan report, the file/directory narrowing in `inspect_protection`, the module list, and the `yes` bool** are all fixed as described. The module list is now the six whose targets actually carry a credential — I checked `Op5Client` and `ElasticClient` too, and their passwords are module-level settings rather than target properties, so your list was exactly right.

**One place I ended up diverging, and it's worth recording.** Dropping `READ_CONTROL` from `protect_directory`'s open is what broke all seven `AclTest`s with `SetSecurityInfo failed: error=5`. Marking a DACL `PROTECTED` converts the inherited ACEs into explicit ones, so the call reads the descriptor before it writes one; without `READ_CONTROL` it fails outright. Your reasoning about which rights the two calls *need* was sound, and the concern behind it was real — so rather than just putting every right back on one handle, ownership and the DACL now take one open each: `WRITE_OWNER`, then `READ_CONTROL | WRITE_DAC`. Taking ownership first is what makes the second open succeed against a folder someone else owns with a DACL that shuts us out, since both those rights come with ownership. That is the same two-step the path-based `SetNamedSecurityInfoW` pair did implicitly, kept explicit and on handles opened with `FILE_FLAG_OPEN_REPARSE_POINT`.

I'd been pushing the Windows-only code unbuilt, which is how two of these got through. `include/win/acl.cpp` and the `WIN32` branch of `upload_staging.cpp` are now cross-compiled against the real Windows headers with mingw-w64 (`-fsyntax-only`) before any push — not a substitute for the Windows CI, but it rules out the compile-level class of error.

---

PR #1516, opened 2026-09-11
feature: add access control gates for data-reading checks

mickem:
## Summary

Adds configurable access control modes to checks whose arguments decide *what data is read* rather than how it is judged. These checks run with agent privileges (SYSTEM/root) and can now be restricted to operator-approved values, preventing callers from reading arbitrary files, WMI classes, performance counters, registry keys, or event logs.

## Key Changes

- **New access policy framework** (`include/check/access_policy.hpp`):
  - Three modes: `any` (default, unrestricted), `allowed` (glob-matched allow list), `predefined` (operator-configured names only)
  - Glob matching with `*` and `?` wildcards, case-insensitive by default
  - Predefined entries resolve in all modes, allowing gradual tightening without rewriting monitoring servers

- **Specialized policy variants**:
  - `path_access_policy.hpp`: For file paths; resolves `..` and symlinks before matching to prevent directory escape
  - `prefix_access_policy.hpp`: For hierarchical namespaces (registry keys, event log channels); supports prefix matching and optional normalization
  - `wql_query.hpp`: Extracts WMI class names from WQL queries for validation; refuses complex forms (ASSOCIATORS OF, REFERENCES OF, etc.) to prevent bypass

- **Integration with checks**:
  - `check_logfile`: `file access` mode with `allowed files` list
  - `check_wmi`: `query access` mode with `allowed classes` and `allowed namespaces` lists
  - `check_pdh` (check_counter): `counter access` mode; predefined set comes from existing `[/settings/system/windows/counters]`
  - `check_files`, `check_single_file`, `check_disk_write`: `file access` mode with `allowed files` list
  - `check_registry_key`, `check_registry_value`: `registry access` mode with `allowed registry keys` list
  - `check_eventlog`: `log access` mode with `allowed logs` list

- **Comprehensive test coverage**:
  - Unit tests for policy parsing, glob matching, mode behavior (`access_policy_test.cpp`)
  - Integration tests for each check module (TypeScript, one-shot client queries)
  - Tests verify refusals don't disclose the allow list and that symlinks/`..` cannot escape restrictions

- **Documentation**:
  - Concepts guide (`docs/docs/concepts/check-access.md`) with examples for each check
  - Security advisory (`docs/security/250-check-access-modes.md`)
  - Upgrade notes (`docs/upgrades/next/check-access-modes.md`)
  - Setup guide integration in `docs/docs/setup/securing.md`

## Notable Implementation Details

- **Fail-closed on configuration error**: Invalid mode values (typos) refuse everything until fixed, not silently fall back to `any`
- **Path resolution before matching**: File paths are canonicalized (symlinks followed, `..` flattened) before glob matching, preventing directory-escape attacks
- **No list disclosure**: Refusal messages name what was rejected and which setting governs it, but never reveal the allow list contents
- **Backward compatible**: All modes default to `any`, preserving existing behavior; upgrading requires no configuration changes
- **Predefined names trusted**: Operator-configured values bypass parsing and validation (e.g., WQL queries are not parsed for class extraction)

https://claude.ai/code/session_01GrJv6RJxMGcFpR8tjE3wh5

### DIFF
--- a/docs/docs/concepts/check-access.md
+++ b/docs/docs/concepts/check-access.md
@@ -16,7 +16,8 @@ the first place*:
 
 That is what those checks are for, and on a host where only the configuration
 decides what runs, it is not a problem. It becomes one where the *caller*
-chooses the argument — NRPE with `allow arguments = true`, or the REST API —
+chooses the argument — NRPE with `allow arguments = true`, or a REST user who
+is not on the no-arguments [`restricted` role](../setup/securing.md#adding-a-dedicated-user) —
 because the agent runs as `SYSTEM` (Windows) or `root`/`nsclient` (Linux) and
 the argument then decides how much of the machine a single check can read back.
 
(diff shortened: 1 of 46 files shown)


mickem:
Three things are blocking this branch. I can diagnose all of them but cannot push the fixes from my session, so recording them here.

### 1. `access_policy_test` fails on every Windows build

`build-server-x64`, `build-server-x86` and `build-x86-static` all fail at **Run tests** on `09e0ab6`, with the same case:

```
[  FAILED  ] path_escape_test.an_element_windows_would_rewrite_is_refused
Expected: false
```

The failing assertion is `EXPECT_FALSE(p.resolve(logs + "\\app.log. ").allowed)`.

**Cause.** In `resolve_physical()` the Windows trailing-space/period guard runs on the *resolved* path:

```cpp
resolved = trim_trailing_separator(to_separators(canonical_path.string()));
...
#ifdef WIN32
    if (has_element_win32_would_trim(resolved)) { ... refuse ... }
#endif
```

Resolution is what destroys the evidence. `weakly_canonical` goes through Win32, which strips the trailing `". "` before the file system sees the name, so `…\logs\app.log. ` resolves to the real name `…\logs\app.log` — inside the allowed directory, with nothing left for the guard to spot — and is accepted. The other two `EXPECT_FALSE` cases resolve to something *outside* `logs`, so the allow list refuses them anyway; that is why only this one assertion fails.

The commit message states the problem correctly ("`logs\.. ` is one odd name to the resolver and `logs\..` to the kernel") — the guard just needs to run **on the token, before `weakly_canonical`**, not only on the result. Note that a naive token-level check also rejects literal `.` and `..` elements, so it should skip those; whether that changes any verdict in the fifteen-spelling traversal table needs a Windows run to confirm.

I have not pushed this fix: the test is `#ifdef WIN32` and I can only build and run on Linux here, so I cannot validate it, and a speculative push costs a ~70-minute cycle.

### 2. DCO is failing

`09e0ab6` and `faf9f82` are authored **and** committed by `Claude <noreply@anthropic.com>` while signed off by `Michael Medin <michael@medin.name>`. The bot needs the sign-off to match author or committer; here it matches neither. The other ten commits on the branch are authored by Michael Medin — these two are the anomaly, and per `CLAUDE.md` only a human may certify the DCO.

Fix: re-apply both commits with `--author="Michael Medin <michael@medin.name>"`. Trees are unchanged.

### 3. `faf9f82` carries `[skip ci]`

Only CodeQL and DCO ran on the current head — no builds, no integration tests, no `get-version`. `CLAUDE.md` warns about exactly this: a skipped workflow leaves its check *Pending*, not *Passed*, so a required check never resolves and the PR cannot be merged. Dropping ` [skip ci]` from that subject in the same rewrite fixes it.

Items 2 and 3 are one operation (cherry-pick both commits onto `7184a8d` with the corrected author, drop the `[skip ci]`, force-with-lease). My sandbox denies history-rewriting git operations, so that one needs a human or a differently-permissioned session. Whoever does it should fetch immediately beforehand and use `--force-with-lease`, since more than one session has been pushing here.

---

Issue #1515, opened 2026-09-11
Vendored replxx brings back the non-free ConvertUTF code


Fantu:
The replxx copy added in #1488 includes `libs/replxx/src/ConvertUTF.{cpp,h}` under Unicode, Inc.'s old license, the same one that was in SimpleIni and was removed in #1294 after I reported it in #1278. Lintian flags it again:

```
E: nscp source: license-problem-old-unicode [LICENSES/LicenseRef-Unicode-ConvertUTF.txt]
E: nscp source: license-problem-old-unicode [libs/replxx/LICENSE.md]
E: nscp source: license-problem-old-unicode [libs/replxx/src/ConvertUTF.cpp]
E: nscp source: license-problem-old-unicode [libs/replxx/src/ConvertUTF.h]
```

The grant is limited to "products supporting the Unicode Standard" and doesn't allow modification, so it isn't DFSG-free. These files are compiled into CommandClient, so unlike the Windows binaries I can't just drop them in a repack, and the package can't go into main.

Two ways out I can see:

- Unicode has relicensed this code: LLVM replaced the header with the current Unicode license (SPDX `Unicode-DFS-2015`) in https://reviews.llvm.org/D66390. The same could be done here, with `REUSE.toml` and `LICENSES/` updated to match.
- Or replace the conversions with `str::utf8`, as #1294 did for SimpleIni.

Background on the Debian side: https://bugs.debian.org/854209

mickem:
Thanks...

I was sceptical to this and went back and forth a few times with the vendoring... 
Original plan was to make it optional and download the library during build and then skip the whole thing on debian to not bother with it...

But claude was insisting that "vendoring was sooo much better especially for the debian usecase"... Unfortunately I had forgotten about the ConvertUTF issue we had a few weeks ago :)

Ahh, well.. I will submit a fix for this...

mickem:
What do you think of this: https://github.com/mickem/nscp/pull/1518

mickem:
This has been fixed now right? If not please reopen.

Issue #1278, opened 2026-05-17
Official Debian packaging


Fantu:
Hi, I took a very quick look to assess the possibility of packaging it for the Debian and derivative repositories.
Debian requires that all dependencies be included in the repository (with rare exceptions); unfortunately, it seems there are missing dependencies (which need to be packaged first), and looking at the GitHub build workflow, it also downloads precompiled binaries (which is unacceptable in Debian).
There are also several parts that require different languages and multiple components, including Rust, which I’m not familiar with and which seems to require a separate procedure.
However, I saw that the Rust part has just been moved to its own separate repository, right? So that should simplify the packaging a bit.

More in details required thing seems:
- tinyxml2: already in debian but with one RC bug about an ABI break in latest version (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1100661), so need to check the compatibility
- [mongoose](https://github.com/cesanta/mongoose) added recently in Debian: https://tracker.debian.org/pkg/mongoose
- [Miniz](https://github.com/richgel999/miniz): missed in Debian but there is an ITP https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1124745
- check_nsclient: if required should be packaged first (not checked rust deps if all already present in Debian)
- about web frontend I have not checked all deps in web/package.json: should check whether the corresponding node-* packages are available in Debian and ensure there are no problematic components; otherwise, we can include all the necessary source files here (tolerated)
- python part: need to check if present in Debian the deps in build/python/requirements.txt

other thing:
- .net: seems needed only in windows or I'm wrong?
- Boost: from a fast look seems all already included with libboost-all-dev
- lua: liblua5.4-dev present and already used also here

mickem:
## General

I guess it would be possible to package a "stripped down" agent only version to reduce effort so for monitoring: you install the packages and if you want admin (UI) features and some exotic module you use the packages built here?


## Tinyxml 
This is used by core and NRDP module and is just an xml reader/writer if there other libs for that I guess one could explore moving to them or look into the ABI breakage, it is not used much so should not be hard either way.

## Mongoose 
This is used only the web server is something I am exploring swapping out as it is not fit for purpose.
The challenge here is that mongoose compiles easily on anything where as many other lib do not.
One candidate for this is to split so we use boost-beast for "modern windows" and linux and mongoose for XP.
In this case it would be swapped out for boost, but I have not explored this yet.

## miniz
[...]

PR #1518, opened 2026-09-12
Replace Unicode-licensed ConvertUTF with DFSG-free implementation

mickem:
Replaces the vendored Unicode, Inc. UTF-8/UTF-32 conversion routines with a new implementation to resolve licensing concerns that keep the Debian package out of main.

## Summary

The upstream replxx library embeds Unicode, Inc.'s 2001-2004 `ConvertUTF.{cpp,h}` files, whose license grants use only "in the creation of products supporting the Unicode Standard" and prohibits modification. This is not DFSG-free and triggers Lintian's `license-problem-old-unicode` flag (same issue as #1278, #1294, #1515).

This change removes those files and replaces them with a new, independently written implementation (`nscp/utf8_conversion.{cpp,hpp}`) that:
- Implements UTF-8 ↔ UTF-32 conversion from the Unicode standard specification (Table 3-7, "Well-Formed UTF-8 Byte Sequences")
- Maintains API compatibility with the replaced code so upstream's `src/conversion.cxx` compiles unchanged
- Fixes two bugs in the original implementation:
  - Rejects `0xED 0x21 0x95` as malformed (original only checked the upper bound of the second byte after 0xED/0xF4 leads)
  - Rejects invalid lead bytes (0xF5+) as `sourceIllegal` rather than `sourceExhausted`
- Includes comprehensive unit tests covering all sequence lengths, edge cases, malformed input, and buffer exhaustion scenarios

## Key Changes

- **Removed**: `libs/replxx/src/ConvertUTF.{cpp,h}` and `LICENSES/LicenseRef-Unicode-ConvertUTF.txt`
- **Added**: `libs/replxx/nscp/utf8_conversion.{cpp,hpp}` with new implementation and `libs/replxx/nscp/utf8_conversion_test.cpp` with 236 lines of test coverage
- **Updated**: `libs/replxx/src/conversion.hxx` to include the new header instead of the old one
- **Updated**: Build configuration (`CMakeLists.txt`, `REUSE.toml`, `tools/license_headers.py`) to reference the new files
- **Updated**: Documentation (`README.nscp.md`, `LICENSE.md`, `THIRD-PARTY-NOTICES.md`) to reflect the change

## Implementation Details

The new implementation:
- Uses explicit sequence length detection and validation helpers (`sequence_length()`, `is_valid_second()`, `is_continuation()`, `is_well_formed()`)
- Decodes/encodes UTF-8 sequences with straightforward bit manipulation rather than lookup tables
- Properly validates all bytes in multi-byte sequences, not just the second byte
- Maintains the same error reporting semantics: `sourceExhausted` for truncated sequences (allowing incremental parsing), `sourceIllegal` for malformed input, `targetExhausted` for buffer overflow
- Supports both strict and lenient conversion modes (lenient allows encoding surrogates; strict rejects them)

https://claude.ai/code/session_01FuTw41rt5QKqtp5g8TP8eq


Fantu:
Thanks, that looks good to me.

One small thing: the commit carries both `Assisted-by:` and `Co-Authored-By: Claude…`, while CLAUDE.md says to use only the former for AI assistance.

mickem:
Good catch — fixed. The commit was amended to drop the `Co-Authored-By:` trailer, keeping `Assisted-by: Claude Code:claude-opus-5` and the `Signed-off-by:`, per the DCO rule in CLAUDE.md that only a human may certify.

The branch was force-pushed, so the head is now `9cb5a2f` rather than `122da75` — the tree is byte-identical, only the commit message changed. CI is re-running on the new head.

---

PR #1522, opened 2026-09-13
nscp test prompt: aligned listings, richer desc, keywords, alias, case-insensitive completion, single quotes

mickem:
## Summary

Follow-up to #1521: a round of usability work on the `nscp test` prompt. Branch is off `main`; seven commits, one per change.

- **Aligned listings.** `queries`, `aliases`, `list` and `plugins` were tab separated, which only lined up while every name was shorter than a tab stop, and a multi-line description spilled into what looked like further rows. Every listing is now a padded table, one line per entry. `list` really lists both queries and aliases now (it used to lose the queries).
- **`desc` shows more.** Parameter defaults, the rest of a multi-line description, a `Default:` line with the command as it runs bare (what `<command> show-default` prints, ready to paste), and for an alias the `Runs:` command line plus that command's parameters. Alias registrations in CheckHelpers and CheckExternalScripts now carry their arguments so the expansion is visible. Fixes the one-character truncation of parameter descriptions.
- **`keywords <query>`** lists a check's filter keywords (and filter functions) with their descriptions, read from the running agent's help payload. For an alias the target's are shown.
- **`alias`** as a shorter spelling of `aliases`; registered as a builtin so help, completion and highlighting know it.
- **Case-insensitive completion.** `load check<tab>` now offers `CheckDisk`, `CheckSystem`, ... and corrects the case of what was typed.
- **Single quotes.** `'...'` takes its content literally (`path='C:\Program Files\app'`); `"..."` keeps its backslash escapes. A single quote only opens a string at the start of an argument or right after its first `=`, so `filter=core='total'` typed bare still reaches the check as written. The prompt has its own tokenizer for this; the shared one, which also reads alias and script definitions from the configuration, is untouched.

```
alias_ok                       Alias for: check_ok message=hi there
alias_with_a_rather_long_name  Alias for: check_timeout timeout=5 command=check_ok

Command:      alias_ok
Runs:         check_ok message=hi there
Description:  Just return OK (anything passed along will be used as a message).
Default:      check_ok "message=No message"
Parameters (of check_ok):
  NAME          DEFAULT     DESCRIPTION
  help                      Show help screen (this screen)
  message       No message  Message to return
```

## Verification

- Console integration suite (`command-client-console.test.ts`): 20 cases, covering alignment across a long alias name, no tabs, the combined `list`, the plugins marker column, `desc` with defaults and alias target, `keywords` for a filter check / a plain check / unknown / bare, `alias` equalling `aliases`, and the four quoting shapes. All pass on Windows.
- Prompt unit tests (`command_client_test`): 41 cases including case-insensitive completion, `keywords` completing like `desc`, and the single-quote highlighting rule. String unit tests: 4 new cases for the prompt tokenizer.

https://claude.ai/code/session_01LSRHcYn26ukbUEjkaVpNDq

### DIFF
--- a/include/client/simple_client.cpp
+++ b/include/client/simple_client.cpp
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2004-2026 Michael Medin
 // SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-only
 
+#include <algorithm>
+#include <boost/algorithm/string.hpp>
 #include <boost/function.hpp>
 #include <client/simple_client.hpp>
 #include <nscapi/macros.hpp>
@@ -14,6 +16,7 @@
 #include <nscapi/protobuf/settings_functions.hpp>
 #include <str/utf8.hpp>
 #include <str/utils.hpp>
+#include <vector>
 
 static void create_registry_query(const nscapi::core_wrapper *core, const std::string &command, const PB::Registry::ItemType &type,
                                   PB::Registry::RegistryResponseMessage &response_message, const bool fetch_all = false) {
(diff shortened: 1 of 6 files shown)


mickem:
CI note on head a3cc2006: `Analyze (actions)` and `Analyze (javascript-typescript)` (GitHub's CodeQL default setup) are red, the other three CodeQL languages are green.

Both jobs scanned every file and reached "Uploading results", then failed with no error text in the log and no annotations. The same two analyses passed on the head pushed two minutes earlier (b4eb4a6f) and are green on `main`, and this push changed only `include/win/pdh/pdh_enumerations.cpp` and a test file, which neither analysis covers. There are no open code-scanning alerts on this head. That points at the upload stage, not this PR.

A re-run is not possible: default-setup runs are "dynamic" and the API refuses to retry them (`This workflow run cannot be retried`). The next code push will trigger a fresh analysis. `main` has no required checks, so the PR remains mergeable once the build matrix is green.

Issue #348, opened 2016-11-20
nscp eventlog command not working


box293:
0.5.0.62 on Window 2008R2.

Following this guide:
https://www.medin.name/blog/2012/03/20/real-time-event-log-monitoring-with-nsclient/

Under "Causing problems" the following command is to be used to insert event log entries:
```
C:\Program Files\NSClient++>nscp eventlog --exec insert-eventlog --source "Application Error" --id 1000 --level error --category 0
```

When I execute the command I get the following message:
```
Usage: nscp eventlog [list|insert] --help
```

So I try looking at the help with the command just provided:
```
C:\Program Files\NSClient++>nscp eventlog insert --help
```

I get the following message saying that **insert** is an invalid option
```
Invalid command line: unrecognised option 'insert'
help                    Show help screen
source=Application Error        source to use
type=                   Event type
level=                  Event level (type)
facility=               Facility/Qualifier
qualifier=              Facility/Qualifier
severity=               Event severity
category=               Event category
customer=               Customer bit 0,1
arguments=              Message arguments (strings)
id=                     Event ID
```

I'm confused

mickem:
in the next verson this works:

```
nscp eventlog insert --source "Windows Error Reporting" --id 1001 --level 4
```

PR #1526, opened 2026-09-14
Claude/fix x 3 c 1 rs69lt

mickem:


mickem:
Rewrote the branch (force-with-lease, content unchanged: same 12 files, +709/−27 against `main`) for two reasons:

- **DCO**: the two commits were authored as `Claude <noreply@anthropic.com>` with the `Signed-off-by` of Michael Medin, which the bot rejects (sign-off must match author or committer). They are now authored and committed as the signer, with the assistance recorded in the `Assisted-by:` trailer, matching every other assisted commit in this repo.
- **Windows CI**: all three Windows `build` jobs failed only in `access_policy_test` (`path_escape_test.an_element_windows_would_rewrite_is_refused`), a file this PR does not touch. The same test is red on `main` at this PR's base (Build release #348) and #1527 fixes it, so the branch is rebased onto `main` at 288bf9a to pick that up. The new upgrade notes moved from the just-filed `docs/upgrades/0.21.0/` back to `docs/upgrades/next/`, where unreleased notes belong.

---

mickem:
Pushed 0a9b963 after a code review of the two commits. What it changes:

- **Windows fallback (no `PROC_THREAD_ATTRIBUTE_HANDLE_LIST`)**: the child-side pipe ends were made inheritable at pipe creation, before the serialising lock was taken, so a spawn already inside `CreateProcess` under the lock could still inherit them. They are now marked immediately before the `CreateProcess` call, under the lock, and un-marked before it is released. With the API present, a per-call failure to build the list now fails the spawn instead of silently falling back to full inheritance. The spawn error is kept in a local rather than round-tripped through `SetLastError`.
- **Unix launcher**: a pipe end landing on fd 0–2 (service started with that descriptor closed) is lifted above stdio first, since `dup2(1, 1)` keeps close-on-exec set. When `close_range()` is unavailable the child now closes exactly the open descriptors listed in `/proc/self/fd` (raw `getdents64`, async-signal-safe) instead of a sweep bounded at 1024/65536 that both leaked higher descriptors and cost one `close()` per possible fd; the sweep stays as the last resort with the bound from the nofile limit.
- **Proxy guard**: dropped the request-marker set (the configured and current values differing already says the request changed the key); the refusal names only options that changed something, so `--host` repeating the configured address is no longer blamed for a proxy-only refusal, and a header that moves both address and proxy names both. Two unit tests pin that; the integration test's allowed cases now also assert the submission was attempted.

Noted but left out of this PR as pre-existing: `do_submit`/`do_metrics` apply header metadata without the host-override guard (reachable by in-process producers only), and `CheckSystemUnix/check_service.cpp` has its own fork/exec helper that gets none of the descriptor hardening.

---

PR #1530, opened 2026-09-14
Ask where the agent gets its configuration, and retire the op5 profile

mickem:
The first page of the Windows installer asked which monitoring vendor was on the other end: *Generic* or *op5*. op5 is deprecated, and the question stopped being the interesting one once the agent could be managed centrally. This replaces it with **Select Management Server** — where does this machine's configuration come from?

| Answer | `MANAGEMENT_SERVER` | What it means |
|---|---|---|
| None | `NONE` (default) | This machine keeps its own configuration, exactly as before |
| NSClient Fleet | `FLEET` | Enroll with a fleet server while installing |
| Web server | `WEB` | An `nsclient.ini` served over HTTP(S) is the configuration |

## What a managed answer does

Both managed modes skip the configuration and module pages, write no local baseline, and put a fresh install on the modern layout. The middle one is the load-bearing part: a value set locally wins over the one the management server sends, so a baseline written at install time would quietly shadow the configuration the operator maintains centrally — the module list included. The only thing written is what points at the management server.

None of the machinery is new. `FLEET_SERVER`, `FLEET_TOKEN`, `FLEET_CA`, `FLEET_BUNDLE_KEY` and `CONFIGURATION_TYPE=https://…` already did all of this on a silent install; the page is a front end for them.

## Compatibility

Silent installs keep working unchanged in the sense that matters: `MANAGEMENT_SERVER` is derived from `FLEET_SERVER`, or from a http(s) `CONFIGURATION_TYPE`, when it is not given, so a deployment script written before this page existed lands in the right mode without being edited.

**One behaviour does change for those scripts, and it is deliberate.** Such an install now writes no local baseline, where before it wrote the generic starting point (`allowed hosts = 127.0.0.1`, a generated password, NRPE, the common check modules) *and* the fleet include. That baseline is exactly what shadows the configuration the fleet server sends, so removing it is the point — but it is a change, and `docs/upgrades/next/010-management-server-page.md` is `action: conditional` and says how to keep a baseline for anyone who wants one.

A derived mode deliberately leaves the on-disk layout alone. Such a command line asked for a fleet server or a configuration url, not for its writable state to move to `%ProgramData%`, and that migration is one-way — so the modern-layout default applies only when the mode was named in `MANAGEMENT_SERVER` or picked on the page. `LAYOUT` always wins where it is given.

## Validation moves forward

A bad url, a missing token, an unparseable bundle key or a CA file that is not on the machine is now reported on the page (`MGMT_ERROR`, shown by a small modal) rather than after Install has been clicked. Enrollment burns a one-time token, so those refusals are worth making before the token is spent. Plain HTTP needs the insecure box ticked in both modes: in one it would send the token in cleartext, in the other it fetches the file that becomes the whole configuration of the agent.

The page and `ScheduleEnrollFleet` share one implementation of those refusals (`fleet_refusal`), so a rule added to one cannot go missing from the other and reappear after the token is gone.

The page owns the properties it collects. Going Back and answering differently takes the earlier answer with it — the fleet server, the token, the CA path, the configuration url, the insecure opt-out and the modern-layout default — because everything downstream keys off those properties rather than off the mode. Where a property means something outside this page (`TLS_CA`, `TLS_VERIFY_MODE`), only a value the page itself set is taken back. The insecure box maps to `FLEET_INSECURE` or to `TLS_VERIFY_MODE=none` depending on the mode, and never across: those are different connections.

`DetectManagement` also notices a host that is already enrolled — the page says so, keeps that identity, and leaves the fields for replacing a bundle key.

## op5

The page's other answer goes with it: the `OP5Monitoring` feature, the `op5.ini` laid beside `nsclient.ini`, four helper scripts, and the two harvested `.wxs` files that fed them. `MONITORING_TOOL=OP5` now installs the generic baseline rather than none at all.

`Op5Client` still ships, moved to *Various client plugins*, and is configured in `nsclient.ini` under `[/settings/op5]` or with `nscp op5 install`. A host that had `OP5Monitoring` but not *Various client plugins* needs `ADDLOCAL=ExtraClientPlugin` on upgrade or it loses the module; the retirement note says so.

One thing worth knowing, found while writing these docs: the `OP5_SERVER`, `OP5_USER`, `OP5_PASSWORD`, `OP5_HOSTGROUPS` and `OP5_CONTACTGROUP` MSI properties **do not work and never have** — `ScheduleWriteConfig` reads them through `KEY_`-prefixed names that no `applyPropertyValue` call ever fills in. That is equally true on `main`, so it is not something this branch broke. The docs here stop claiming otherwise and the rows are out of the options table; whether to make them work or delete the dead code is a separate call, and I have left the code as it is.

## Testing

`management-web.yaml` installs with `MANAGEMENT_SERVER=WEB` and asserts the url reaches `boot.ini`, nothing local is written over it, and the install lands on the modern layout. `monitoring-tool-op5.yaml` pins that an un-updated op5 command line gets the generic configuration, with the four retired scripts confirmed absent. `boot.ini-rest-config-verified.yaml` already covered the derived case and now pins the other half without a line changing — managed mode, layout untouched.

The full MSI suite passes on Windows, all 22 cases.

The dialog itself cannot be driven by the MSI suite. Worth a manual pass before merging:

- Welcome → the page → each of the three answers; None reaches the configuration and module pages, the other two go straight to Ready to install.
- A fleet answer with an empty token, an `http://` url, and a mistyped bundle key each stay on the page with a readable message, whole — the error dialog was resized for the longest of them.
- Web, tick insecure, type a CA path, then Back and answer None: the finished install should have no `verify mode`, no `ca` and no url in `boot.ini`, and stay on the legacy layout.
- An upgrade of an enrolled host preselects NSClient Fleet and names the server.

Upgrade notes are under `docs/upgrades/next/`.

https://claude.ai/code/session_018swDdX2Bi6pVrgR1EvYMrK


mickem:
**CI: `build-x86-static / integration-tests` is red, and it is not this branch's.**

One test failed, in [job 105726638719](https://github.com/mickem/nscp/actions/runs/35377121154/job/105726638719):

```
● nscp test console › keeps running commands after a module reload
    expect(received).toBe(expected) // Object.is equality
    Expected: 0
    Received: 3221225477
      at command-client-console.test.ts:361:24
```

`3221225477` is `0xC0000005` — an access violation. The console did the work: `after-reload` is in the output and `timedOut` is false. The process then died on the way out, so only the `exitCode` assertion failed. Everything else in that suite passed (920 of 921).

Why it is not this PR's:

- The test and the code under it (`modules/CommandClient`, the reload-lifetime fix in 29ecfba) are on `main` and this branch does not touch either. The only non-installer source change here is one static string in `installer_feature_hint()` (`service/plugins/plugin_manager.cpp`) — moving `Op5Client` from the retired *OP5 Monitoring system* feature to *Various client plugins*. It is a map lookup used to compose an error hint; it cannot reach a console reload.
- The same job is green on `main` at this PR's base commit `949a90a` ([job 105666042130](https://github.com/mickem/nscp/actions/runs/35355406123)).
- `build-server-x64 / integration-tests` and `build-server-x86 / integration-tests` both passed on this same head, including the MSI installer suite that covers what this PR actually changes (`management-web.yaml`, `monitoring-tool-op5.yaml`).

So it is an intermittent pre-existing crash that happens to show on the static x86 build. There is no fix on `main` to port into this branch, and I have not widened this PR to chase it. I have re-run the failed job once to confirm it does not reproduce.

Worth noting for its own issue rather than this one: 29ecfba fixed the *completion-hook* half of the reload lifetime problem (the editor's captured `cli_client` pointer). This failure is on the piped path, where no editor exists, and the crash is at teardown after `reload` — so there is plausibly a second lifetime issue left in that area, visible only when the timing lines up.

---

PR #1531, opened 2026-09-14
Fix the ten highest-severity threading and ordering bugs from the concurrency audit

mickem:
Acts on the top ten findings of the whole-tree concurrency audit, one commit each. Every finding was re-verified against current `main` before being fixed.

Since the first round of review the branch has also grown an integration suite for the threading contract itself, plus the fixes that suite uncovered — see **Threading tests** and **PythonScript** below. Rebased onto `cf4521a`.

## The common root

A settings reload calls `loadModuleEx(reloadStart)` on the live module. Unload closes the door and waits for the calls already inside; load never did. Every handler therefore kept running straight through the settings the reload was rewriting, which is where most of the module-level races come from. The second commit gives load the same barrier, which is the single highest-value change here.

## What each commit fixes

| Commit | Problem |
|---|---|
| collector published atomically | A reload `reset()` the collector while checks were inside it, destroying a locked `shared_mutex` and its buffers under an NRPE thread. Heap corruption with no usable stack. |
| reload dispatch barrier | The root cause above. New dispatches wait while `loadModuleEx` runs; the reload waits for the calls already inside. |
| refuse self-unload | `POST /api/v2/modules/WEBServer/commands/unload` ran on the web server's own thread, joined that thread with itself, then unmapped the library while the thread was still executing inside it. |
| scheduler watchdog vs `stop()` | The watchdog cleared `stop_requested_` from inside the pool while `stop()` was joining, so workers looped forever and shutdown hung. Auto-scaling is on by default. |
| settings lock over I/O | `instance_mutex_` was held across an untimed HTTP configuration download, so every `get_string` on every thread threw "Failed to get mutex" for the duration. |
| refused unload | After `unload_plugin()` deliberately refused, the destructor called `dlclose` anyway and `remove_plugin` left the module half removed. |
| lifecycle serialisation | `find_duplicate` and the append that follows it were not atomic, so two loaders could both pass and create two instances of one module. |
| socket server self-stop | Reloading a listener from inside a check that listener is serving joined the caller with itself. |
| python registry under the GIL | The function maps were written from Python under the GIL but read without it, and the iterator was kept across the acquisition that followed. |
| lua nested dispatch | A script querying a command its own module serves asked for a writer-preferring `shared_mutex` twice on one thread and deadlocked against a waiting writer, permanently. |
| lua manager published atomically | The check threads copied the `scripts_` shared_ptr plainly while load and unload reset the same member — a race on the pointer itself, not merely on what it points at. Now `atomic_store` / `atomic_load` / `atomic_exchange`, like the collectors. |
| no raw collector pointer | `CheckDisk::get_collector()` returned `get_collector_ptr().get()`, whose owning temporary dies at the end of the caller's expression. Test-only, but nothing in the signature said so. Removed. |
| client connect deadline | The configured timeout covered only the exchange, not the connect or the TLS handshake - the two steps most likely to hang. |
| `check_ping` re-arm | One stranger's echo reply consumed the single outstanding receive, so our own reply was never read and the host was reported as losing packets. |

## Threading tests

`tests/plugin-threading.test.ts` asserts the guarantees the core gives every module. Nothing in the suite noticed them before: every existing check test issues one command at a time, so a core that serialised every dispatch under one lock passed all of them.

It drives NRPE rather than REST, because the web server runs its handlers on a single `ioc_.run()` thread and concurrent REST requests would measure the HTTP layer instead of the dispatch path. The client is `nscp nrpe`, the alias built into the same binary, since `check_nscp_nrpe` is a build-tree artifact and is not packaged.

| Case | Guarantee |
|---|---|
| four callers of a blocking script | several callers are inside one module at once |
| slow module + `check_ok` | a blocked module does not stall an unrelated one |
| `check_multi` | a module dispatched into by itself does not deadlock |
| `lua_nested` / `lua_deep` | the same through the core, one and two levels deep |
| `lua_busy` x4 | concurrent scripted checks do not cross state |
| python boot / `py_native` / `py_nested` / `py_sleep` | see below |
| reload under load | in-flight checks survive `loadModuleEx(reloadStart)` |

Overlap is recorded by the checks themselves (a marker on entry and on exit, replayed from a trace file) rather than inferred from the client's clock, so a loaded runner cannot turn the result into a coin flip. Validated negatively: with `thread pool = 1` the concurrency cases fail as they should — peak overlap 1 against an expected > 1, and zero checks served while another module was blocked.

## PythonScript

Writing the tests above turned up a bug that killed the agent at boot for anyone with a Python script configured: `loadModuleEx` ran `settings.notify()` — which constructs every configured `python_script`, and that constructor takes the GIL — before `python_script::init()` ran `Py_Initialize()`. It reproduced against the shipped `sample.py`, and nothing caught it because no test enabled the module with a script.

**That fix is now on `main`** as `9e0da11`, arrived at independently and by the same route (defer the script construction into a queue, initialise the interpreter after `notify()`). This branch was rebased onto it and carries no production change to PythonScript any more — only the tests, which `main` does not have:

- **boot with a script configured** — the case that would have caught the segfault.
- **`py_native`** — imports `ctypes`. This guards the *other* half of the ordering, which is easy to miss: `init()` must also run **after** `notify()`, because `notify()` is what assigns `python lib`, and an empty one skips the `RTLD_GLOBAL` dlopen. Get that wrong and the agent boots perfectly while every C extension import dies with `undefined symbol: PyTuple_Type`, taking protobuf and every bundled script with it. A pure-Python test never notices.
- **`py_nested` / `py_sleep`** — the threading contract, below.

**Threading, for the record:** PythonScript does *not* have the Lua defect below. Both modules release the interpreter lock around core calls, but `PyGILState_Ensure` gives every OS thread its own `PyThreadState`, so the releasing thread's frame stack is private. 90 concurrent `py_nested` calls returned 90 correct results with the agent still serving.

`NSCP.sleep` genuinely yields, too: three concurrent checks that each sleep 2 seconds finish in ~2.0 s wall, not 6, and an unrelated module answers while all three are parked. The wall-clock bound is what proves it — CPython switches threads every few milliseconds regardless, so even CPU-bound Python shows several checks "inside" at once. Swapping the sleep for ~2 s of pure bytecode makes the same three calls take 6.2 s and fails the bound, which is what a GIL held across the sleep would look like.

## Known defect left open: concurrent Lua that calls into the core

One test is checked in **skipped**, with the diagnosis in a comment, rather than enabled: two callers of a Lua script that calls back into the core corrupt the interpreter and kill the agent (SIGSEGV, or a Lua PANIC to abort).

`lua_core.hpp:75` hands every invocation `information->user_data.L` — one `lua_State` per *script*, shared by all concurrent invocations — and `lua_script.cpp:84` drops the Lua lock around core calls. So while one thread is in the core, a second runs that same `lua_State`. It reproduces on an unmodified `main`, so it predates this branch, and fixing it means giving each invocation its own execution state (a `lua_newthread` coroutine) or holding the lock across core calls — a change to the Lua threading model rather than a tweak. Left for its own PR.

## Verification

On the rebased tree: clean build (723/723), the full unit suite passes **102/102**, the threading suite is 11 passing / 1 skipped, and `main`'s own `pythonscript-metrics` suite passes 2/2 (checked because this branch takes `main`'s PythonScript sources wholesale).

One change was reverted during this work rather than shipped. An earlier version deferred *every* instant reload requested from inside a module to the scheduler. That broke `lua_uptime_test` and `lua_ext_script_test`, which reload from a script and immediately read the new configuration back - a real contract, not a flaky test. The narrower guard in `server::stop()` covers the actual hazard without changing reload semantics, and both tests pass.

Windows-only code (`CheckSystem`, `CheckEventLog`) could not be compiled here and was reviewed rather than built; everything else was compiled. Windows is covered by CI, which ran the threading suite on x64, x86 and x86-static before the rebase.

## Notes

- `docs/upgrades/next/` carries a note for the operator-visible changes: reloads now wait for running checks, a module cannot unload or restart itself from inside a request, HTTP settings downloads have a 30 second deadline, and the submission clients apply their timeout to connect and handshake.
- No security notice: these are robustness fixes for races triggered by the operator's own reload or by an already-privileged API call, not remotely reachable vulnerabilities. Worth a second opinion if you disagree.
- Unrelated flake seen on this branch, **not fixed here**: `onboarding_test` / `SyncCertHostile.ExpiryBoundariesAroundTheRenewalThreshold`. `make_cert_expiring_in(0)` stamps `notAfter` at exactly now, and `ASN1_TIME` has one-second granularity, so generating the certificate across a second boundary reads back as `-1` instead of `0`. Reproduced 1 run in 400 locally. A one-line test fix (use `make_cert_expiring_in_seconds(1800)`) gives 0 in 4000, but it is unrelated code and belongs in its own change.
- The remaining audit findings (the other ~50) are untouched.

https://claude.ai/code/session_01BZFiRq7AaC9mTk7nZ53sri


mickem:
Adding one more commit, and recording an intermittent failure that is worth knowing about when reviewing this.

## The failure

On d9f72cb, `build-server-x64 / integration-tests` hung. The REST scenarios passed in ~10m, then `Run installer tests` started at 15:08:46 and never finished; the job was killed 44m into that step. A re-run of the same job on the same commit passed, with that step taking 4m09s.

It is not runner slowness. On main (run 34854081094), same runner pool, same hour, that step took 4m and the whole job 25m, with near-identical timings for every earlier step. The step only runs on x64, so the two green x86 integration jobs never exercised it.

The mechanism fits a service that will not stop: `Product.wxs` has `<ServiceControl Start="install" Stop="both" Wait="yes">`, so msiexec waits on the service, and `tests/msi/helpers.py` kills each msiexec at 120s. Roughly 22 test cases at 120s is the 44 minutes.

That job's logs are permanently unavailable (404), so this could not be read directly, and the hang did not reproduce.

## The commit

Whether or not it caused that hang, there is a real deadlock in the shutdown path that the earlier scheduler commit left open, and this PR should not ship it half fixed.

`boost::thread_group::join_all` holds the group's shared lock for the whole join while `create_thread` wants it exclusively. The watchdog scales the pool up from inside the pool, so a watchdog that passed the `running_` check just as `stop()` began could sit in `create_thread` waiting for a lock that `join_all` only releases once every thread has been joined, including that watchdog. `stop()` then never returns and the process never exits.

The earlier commit closed the flag half of that race and left this half open, because it released `pool_mutex_` before the join. This one holds the lock across the interrupt and the join so nothing can reach `create_thread` while a join is in flight, and gives the watchdog a non-blocking entry point that skips the scale-up when the lock is held, since blocking there would mean waiting for the thread that is waiting to join it.

Verified with a clean build and 102/102 unit tests, plus five service start/stop cycles on Linux with modules loaded, each shutting down in about a second. I could not reproduce the Windows hang locally.

---

PR #1534, opened 2026-09-14
Second undefined-behaviour sweep: the crashes left in the script hosts and the Windows collectors

mickem:
The C++ tree was read again after the audit that shipped in 0.20.0 (#1505). This branch fixes the crashes that sweep left behind: the ones rated High, one per commit.

None is reachable by an unauthenticated peer. Each needs either a script the operator installed or a configuration value they set.

## What is fixed

| Commit | Defect |
|---|---|
| `fix(lua)` | A script's error text was used as `luaL_error`'s **format string**, so a `%` in anything the script passed in read a wild pointer. |
| `fix(lua)` | A non-numeric argument threw `bad_lexical_cast` out of a `lua_CFunction`, unwinding through Lua's C frames. |
| `fix(python)` | A string Python cannot encode as UTF-8 (a lone surrogate) reached `PyBytes_AsString(NULL)`. |
| `fix(python)` | `PyRun_File` was handed the `FILE*` from `_Py_fopen_obj` unchecked, and the stream was never closed. |
| `fix(CheckSystem)` | An rrd counter with `buffer size = 0` built a buffer holding nothing; the collector thread called `back()` on it once a second. |
| `fix(CheckEventLog)` | `check_modern` closed event handles the filter object already owned — on the default `scan-range=-24h` path, every run. |

Two High findings from the sweep are not in the diff, because main fixed them first:

- The CommandClient reload use-after-free, fixed by #1521 before this branch opened.
- The PythonScript interpreter-ordering crash, which made any agent with a Python script configured die as it started. This branch carried its own fix for it (44f1692); #1535 landed an independent one while the PR was open, so the merge takes main's version and drops this one. Main's is the better fix: it also clears the recorded scripts, so a reload does not load every script twice. The security notice still describes the crash, since that is where an operator reads about it.

## Verification

- `libs/lua_nscp/lua_cpp_test.cpp` is new and covers both Lua fixes. Each test fails without its fix; the format-string one segfaults in `lua_pushvfstring` exactly as reported.
- The Python fixes were checked end to end on Linux with Python 3.12: a check returning `'name=bad\udcfftail'` segfaults in `PyBytes_AsString` before the change and answers `name=bad?tail` after.
- `pdh_objects_test.cpp` covers the buffer-size guard; `SetBufferSizeInvalidResetsToZero` encoded the old behaviour and is replaced.
- Full build plus all 102 unit tests pass under AddressSanitizer and UndefinedBehaviorSanitizer, before and after the merge with main.
- The two Windows fixes compile only on Windows, so CI is their first real check. Both Windows builds and all three Windows integration-test runs passed on the pre-merge head.

## Behaviour changes

A counter whose `buffer size` is zero or unparseable is now skipped and named in the log rather than loaded with an empty buffer. A Lua binding given a word where it expects a number reads `0` instead of terminating the agent. Both are in the upgrade note; the branch also carries the matching security notice.

https://claude.ai/code/session_01NQjHqfZZAzdTvsRTPMWhW1


mickem:
Both `build-redhat / integration-tests` jobs (rocky 10, x64 and arm64) failed on `77704bb`, and the failure is not this PR's.

They died in the dependency install, before checkout, build or any test:

```
Error:
 Problem: cannot install the best candidate for the job
  - nothing provides openssl-libs(aarch-64) = 1:3.5.8-1.el10_2 needed by openssl-devel-1:3.5.8-1.el10_2.aarch64 from appstream
##[error]Failed after 3 attempts: dnf install -y --allowerasing ... openssl openssl-devel ...
```

Rocky 10's appstream is offering `openssl-devel 3.5.8-1.el10_2` while the repo still carries `openssl-libs 3.5.5-6.el10_2`, so the toolchain install cannot resolve. What points away from this branch:

- It hit x64 and arm64 identically, and names RPM packages. This PR's diff touches C++ sources, one CMake test target and docs; no packaging, no workflow, no dependency.
- The same job passed on `main` at 13:20 UTC and on this PR's own pre-merge head at 00:20 UTC. The mirrors went inconsistent between then and 14:14 UTC.
- The `build-rpm` jobs on the same run, which install a smaller set, passed.

There is no fix to port: it clears when Rocky's mirrors sync. Pinning or adding `--nobest` to the install line would be a workflow change to paper over a transient repo state, which does not belong in this branch. I am holding the one allowed re-run until the mirrors have had time to settle, and will re-run then; if it persists, it will start failing on `main` too and wants a fix of its own.

Everything else on this head is green or still running, including both Windows builds.

---

mickem:
Closing the loop on the re-run I said I would hold: I spent it at 15:53 UTC, 100 minutes after the first failure, and both RedHat jobs failed identically.

```
nothing provides openssl-libs(aarch-64) = 1:3.5.8-1.el10_2 needed by openssl-devel-1:3.5.8-1.el10_2.aarch64 from appstream
```

Reproducing unchanged after that long makes it a persistent Rocky 10 repo skew rather than a passing mirror glitch, so it will start failing on `main` at the next push there. It is not this branch's, and I am not adding a CI workaround to a branch of crash fixes: it would turn this PR green while `main` stays broken.

The fix belongs in its own change, and is one word. `.github/workflows/integration-tests-linux.yml:96`:

```diff
-          retry dnf install -y --allowerasing coreutils bash file findutils git boost-devel ...
+          retry dnf install -y --allowerasing --nobest coreutils bash file findutils git boost-devel ...
```

`--nobest` lets dnf settle for `openssl-devel 3.5.5`, which matches the `openssl-libs` the image actually has, instead of insisting on the 3.5.8 the repo advertises without its runtime. `build-redhat.yml:85` installs a smaller set and is not hitting it today, but has the same exposure. Happy to open that PR if you want it.

Everything else on `77704bb` is green: all three Windows builds and all three Windows integration-test runs, both Debian integration-test runs, all four RPM builds, web, CodeQL, DCO and the docs checks.

---

PR #1538, opened 2026-09-15
feature: OpenMetrics labels for per-instance metrics

mickem:
Per-instance metrics (per-core, per-NIC, per-drive, per-process) are now published as one OpenMetrics family with a label, rather than one family per instance. This enables `sum by (core)` aggregations, Grafana variables bound to labels, and stable family names across hosts with different hardware configurations.

> Rebased on the metrics-metadata work (#1534), which landed while this was open and had already built the **renderer** half of labels — `dims_of()`, `alias` handling, escaping and the family/series bookkeeping. No producer set a dimension, so the endpoint still named a family per core, per NIC and per drive. This PR is the producer half, plus the builder surface it needs.

## Summary

Metrics measured once per core, NIC, drive, thermal zone, battery or process now use OpenMetrics labels to distinguish instances, rather than embedding the instance in the family name. The metric key — read by the JSON endpoints, the web UI, Graphite, collectd and Python `submit_metrics` — is unchanged; labels are additive.

**Before:**
```
system_cpu_core_0_idle_percent 93
system_cpu_core_1_idle_percent 91
system_cpu_total_idle_percent 95
```

**After:**
```
# HELP system_cpu_idle_percent Share of CPU time spent idle
# TYPE system_cpu_idle_percent gauge
# UNIT system_cpu_idle_percent percent
system_cpu_idle_percent{core="0"} 93
system_cpu_idle_percent{core="1"} 91
system_cpu_idle_percent{core="total"} 95
```

## Key changes

- **`instance()`, `label()` and `key()` on the metric builder.** `instance()` composes the key exactly as the concatenation it replaces did, so every consumer that reads keys sees a byte-identical snapshot; `label()` records the dimension and the family name, which is what the renderer groups on. `key()` is for a producer whose historical key does not put the instance first — a PDH counter puts it last.
- **`core_label()`.** Windows spells a core `core 0` and Linux `core_0`. That difference is a fact about keys dashboards already read, and it must not reach the label, where it would make one core look like two across a fleet. Both reduce to the bare `0`.
- **Producer sweep.** CPU load, network, temperature, CPU frequency, battery and process history on both platforms; CheckDisk io and free space; PDH counters configured with instances — which publish the instance as a key *suffix*, so `pdh_thread` now records the counter/instance split where it is still known rather than trying to recover it from a key whose counter name can itself contain dots.
- **Python `labels`.** A `fetch_metrics` callback can add `"labels": {"queue": "inbound"}` to the dict form. A scalar still means an unlabelled gauge.
- **Info families split per instance.** A string metric becomes a label of its bundle's `_info` family, and the instance labels decide which series it lands on — so one NIC's link state, MAC address and speed share a line and the next NIC's get their own, instead of every NIC's strings piling onto one series under label names like `eth0_status`.

## Compatibility

`/api/v2/metrics`, `/metrics`, the web UI dashboard, Graphite, collectd and Python `submit_metrics` are byte for byte unchanged — pinned by a test asserting the flat key set, and by `web/src/metric_parser.test.ts` staying untouched and green. `add_metric()` is unchanged, so an out-of-tree module compiles and renders as before, and `openmetrics format = legacy` still reproduces the pre-renderer body byte for byte (a labelled metric keeps its key, so the legacy body is identical whether or not a producer was swept).

An upgrade note under `docs/upgrades/next/` covers the family renames and the `core="total"` caveat for `sum without (core)`.

## Testing

Green on every platform CI builds. Locally on Linux: 101/101 unit test binaries, 26 builder tests, 54 renderer tests, 20 metrics integration tests and the PythonScript metrics suite, plus the checkdisk / checksystem / collectd / graphite suites and the web UI parser tests — and a real exposition captured and inspected end to end.

The Windows-only sources (`CheckSystem.cpp`, `pdh_thread.*` and the five `check_*.cpp` structs) could not be compiled in the dev container, so they were type-checked by hand against each terminator; the three Windows builds and their integration-test jobs then confirmed them.

https://claude.ai/code/session_01MAVKRH8c4xAvUiEErY4iaK


mickem:
CI note: `build-redhat / build-rpm (rocky, 10, python312, …)` is red on both x64 and arm64 for head `2c7a7bd`. **This is not this PR's failure**, and I am not fixing it here.

It fails in the `dnf install` step, before anything is configured or compiled:

```
Problem: package openssl-1:3.5.8-1.el10_2.x86_64 from baseos requires
         openssl-libs(x86-64) = 1:3.5.8-1.el10_2, but none of the providers can be installed
  - cannot install both openssl-libs-1:3.5.5-6.el10_2 from baseos
                    and openssl-libs-1:3.5.8-1.el10_2 from @System
  - package openssl-devel-1:3.5.5-6.el10_2 from appstream requires
    openssl-libs(x86-64) = 1:3.5.5-6.el10_2
```

Rocky 10's `baseos` has moved `openssl`/`openssl-libs` to 3.5.8 while `appstream` still ships `openssl-devel` 3.5.5-6, and the container image already carries `openssl-libs` 3.5.8. There is no set of packages satisfying both, so dnf gives up — the workflow's own three-attempt retry reproduced it identically on both architectures.

Why it is not this PR's:

* The same job **passed on this PR's previous head** `ae29b3b` at 12:46–12:51 UTC today, and passed on rocky 9 in this very run.
* This push touches no workflow, no spec file and no dependency — only C++ sources, docs, and two removed lines in `tests/CMakeLists.txt`, all of which run long after `dnf`.
* Rocky 9 (both arches), Debian (both arches), all three Windows builds and every integration-test job are green on this head.

No fix for it exists in the repo or in any open PR, so there is nothing to port. The remedy is either waiting for `appstream` to publish `openssl-devel` 3.5.8, or letting dnf settle for a matched older set — in `.github/workflows/build-redhat.yml`:

```diff
-        retry dnf install -y --allowerasing coreutils bash file findutils git boost-devel …
+        retry dnf install -y --allowerasing --nobest coreutils bash file findutils git boost-devel …
```

`--nobest` lets dnf pair `openssl` 3.5.5-6 with the `openssl-devel` 3.5.5-6 that appstream actually has, instead of insisting on the newest `openssl` and finding its `-devel` missing. That is a CI change unrelated to this diff, so I have left it as a proposal rather than widening the PR — happy to push it here or open it separately if you would rather have it now.

I am spending the one allowed re-run on the failed jobs; the repository metadata may well have caught up by now. If it fails the same way I will not re-run again, and will keep watching the PR.

---

PR #1544, opened 2026-09-17
build: share the common plugin code instead of copying it into every DLL

mickem:
40-odd module DLLs each compiled their own copy of the same support code. `plugin_api` / `nscp_protobuf` / `nscp_where_filter` only ever annotated part of their surface with `__declspec(dllexport)` — `nscapi_helper.hpp`, `settings/helper.hpp`, `nscapi_program_options.hpp` and `str/utf8.hpp` carry no export macro at all — so that code was unreachable from a module and each module compiled the source again. **Three of the five files in `NSCP_DEF_PLUGIN_CPP` were already inside `plugin_api`, which all 43 modules link.**

The worst single offender: `include/json/use_json.cpp` is one line — `#include <boost/json/src.hpp>`, i.e. the whole of Boost.JSON — compiled into 12 separate binaries at ~660 KB of object code each.

## What changed

`WINDOWS_EXPORT_ALL_SYMBOLS` on the three libraries that need it, then the shared sources moved out of the modules:

| Library | Now holds | Was compiled privately by |
|---|---|---|
| `plugin_api` (existing) | `nscapi_program_options.cpp`, `nscapi_metrics_helper.cpp`, `metrics_store_map.cpp` | 43 / 3 / 4 targets |
| `nscp_net` (new) | `socket_helpers.cpp`, `allowed_hosts.cpp` | 21 targets, ~520 KB each |
| `nscp_client` (new) | `command_line_parser.cpp` | 12 targets, ~580 KB each |
| `nscp_json` (new) | Boost.JSON, via `BOOST_JSON_DYN_LINK` | 12 targets, ~660 KB each |

`NSCP_DEF_PLUGIN_CPP` drops to the one file that genuinely cannot be shared.

Deliberately still per-module, because they read `nscapi::plugin_singleton` — the global each module defines for itself in its generated `module.cpp`:

- `nscapi_plugin_impl.cpp`
- `simple_client.cpp`, which logs through the `NSC_LOG_*` macros (`GET_CORE()`). Sharing it was tried; the link says no. The reason is in the comments so nobody retries it.
- `CheckPowershell` keeps its own copies of `utf8.cpp` and `nscapi_helper.cpp`: it is a `/clr` thin plugin with an empty `target_link_libraries`.

## Measured size effect

CI artifact totals, against the base commit. `prev` is this branch before the latest commit narrowed `WINDOWS_EXPORT_ALL_SYMBOLS` from every shared library to the three that need it:

| artifact | base | prev | now | now vs base |
|---|---|---|---|---|
| ubuntu-24.04-x64 | 32,277,294 | 25,879,486 | 25,869,308 | **−19.9%** |
| rocky-10-x64 | 17,641,407 | 15,352,359 | 15,350,965 | **−13.0%** |
| rocky-9-x64 | 16,860,276 | 14,809,886 | 14,809,616 | **−12.2%** |
| Win32 | 136,285,178 | 133,468,051 | 128,528,477 | **−5.7%** |
| x64 | 165,553,965 | 162,624,489 | 156,640,285 | **−5.4%** |
| Win32-legacy-xp | 184,660,640 | 186,027,483 | 186,032,350 | **+0.7%** |

Narrowing export-all is worth **−6.0 MB on x64 and −4.9 MB on Win32** on its own. Applying it to libraries that already declare what is public costs real size, because an exported symbol is a root the linker may not discard, so `/OPT:REF` stops pruning unreached code anywhere in that DLL.

**Two caveats on reading this table.** These are whole-artifact totals, which include the separate symbols zip; on the previous commit the maintainer measured the shipped zip at 263 MB → 264 MB and the MSI at 71 MB → 72 MB while the artifact showed −1.8%, i.e. the artifact delta was PDBs, not binaries. The shipped-size effect of the latest commit has not been confirmed yet. And an earlier revision of this description extrapolated a Linux-only measurement into a claim that the Windows saving "should be larger"; that was wrong and is retracted.

The static XP build is **unchanged by both fixes** and still +1.37 MB against base. The cause is not what an earlier commit message guessed: `CollectdClient` and `SyslogClient` used to compile their own SSL-free copy of `socket_helpers.cpp`, and now link the shared one, which `nscp_net` always builds with SSL on. Where OpenSSL is a DLL that costs nothing; in the static build it links OpenSSL into both modules. Whether to accept that, let those two keep a private copy, or split the SSL parts out of `nscp_net` is open.

One mechanism was tested and **disproved**: that MSVC's `/OPT:REF` had already pruned the duplicated dead code. Rebuilding both sides on Linux with `-ffunction-sections -fdata-sections -Wl,--gc-sections` keeps the saving nearly intact (−19.2% vs −19.6%, with the baseline itself only 1.0% smaller).

`USE_SSL` is also `PRIVATE` again on `nscp_net`. It had been made `PUBLIC` on the stated grounds that it changes the layout of `socket_helpers::connection_info`. That was wrong: every `#ifdef USE_SSL` in `socket_helpers.hpp` guards an include, a free function or a member function, and not one guards a data member, so the layout is identical either way and consumers can mix the settings safely — as they always did. The claimed ODR bug did not exist. Reverting it changed the size by 5 KB, so it was a correctness cleanup, not the size fix it was predicted to be.

## What the first CI run caught

The branch was prepared without MSVC available, so the first Windows run was the first real link — 19 projects, 60 errors, in two ways Linux structurally cannot show.

**`LNK2001` on `connection_info::backlog_default`, 17 projects.** A class static is a *data* symbol. Its only definition now lives in `nscp_net`, and reaching a data symbol across a DLL boundary needs `__declspec(dllimport)` on the declaration — being in the generated `.def` is not enough. Fixed by making it `static constexpr`: C++17 makes that implicitly inline, every translation unit gets the value directly, and no symbol crosses the boundary.

**`LNK2005` × 23 on `use_json.obj`, 2 service tests.** They compiled `use_json.cpp` while also linking `nscp_json` through `${EXTRA_LIBS}` (which carries `onboarding`, whose objects reference Boost.JSON). Every test target now links `nscp_json` instead of compiling its own copy. ELF resolves such a conflict silently by preferring the local definition, which is why it only appeared on Windows.

The original audit checked only for *missing* symbols, which is one-directional. There is now also a check comparing each binary's own defined symbols against those exported by the shared libraries it links. It reports only `boost_asio_detail_posix_thread_function` (POSIX-only) and `b64_encode`/`b64_decode` shared by `WEBServer` and `nscp_mongoose`, which predates this branch and which CI links cleanly.

## Verification

- All 28 checks green on the current head, including both Windows dynamic builds, the XP static build, and the Windows integration tests on x64 and x86.
- 102/102 unit tests pass.
- Full integration suite: 782 passed / 10 failed — identical to the baseline build, same ten cases (IPv6 unavailable in the container, `PythonScript` not built without Boost.Python).
- No module, library or executable gained an unresolved symbol against the baseline.
- The three new DLLs are added to `installers/installer-NSCP/RUNTIME_DLL.wxs`; an upgrade note is under `docs/upgrades/next/` (`python3 docs/hooks/notes.py --check` passes).

**Checked and not worth doing:** converting `perfconfig_parser` / `expression_parser` to shared libraries. Their real per-module footprint measures ~21 KB — the linker already discards the unused objects.

https://claude.ai/code/session_01547nfMYxF8CB1or4bsYMqN


Fantu:
I built this branch with my (still unofficial) Debian packaging to see what it does to a real package. Agent-only build on sid: WEBServer and DotnetPlugins off, 34 modules, the private libraries in `/usr/lib/nsclient/` with a `$ORIGIN/..` RUNPATH.

Measured against the same packaging on 0.21.0:

| | 0.21.0 | this branch | |
|---|---|---|---|
| `nsclient_*.deb` | 11.59 MB | 9.21 MB | **-20.6%** |
| Installed-Size | 70,354 kB | 55,056 kB | -21.7% |
| the 34 modules | 61.5 MB | 46.3 MB | -24.7% |

The new libraries are 334 KB (`nscp_client`), 515 KB (`nscp_json`) and 372 KB (`nscp_net`); `plugin_api` grows from 576 to 748 KB. These are installed sizes of stripped binaries, with no debug symbols anywhere in the number, so they are not directly comparable to the CI artifact totals in your table.

Nothing had to change in the packaging: the new libraries are picked up by the glob that already ships the other three, the modules resolve them through the existing RUNPATH, and the binary dependencies are byte for byte identical. 90/90 tests pass and the build is a little faster (1951s against 2171s). Installed in a clean chroot, `ldd` reports no missing symbol for the binary, the modules or the libraries, and the checks I ran answer normally.

Two new lintian remarks, both expected and both harmless: `hardening-no-fortify-functions` on `libCauseCrashes.so` and `libSamplePluginSimple.so`, which shrink from 350 to 109 KB and from 338 to 84 KB and no longer contain any fortifiable call.

One question, from the packaging side. `libs/json` compiles `boost/json/src.hpp` into `nscp_json`, while distributions ship Boost.JSON as a library of their own: Debian sid has `libboost_json.so.1.90.0` in `libboost-json1.90.0`, and the other distributions you build for package it too. Linking the system one where it exists would save that copy and is what a distribution would rather see; keeping the current behaviour on Windows and in the static build. Would you take a patch that does this behind a CMake option, or do you prefer one copy everywhere?

mickem:
CauseCrashes should not be built and definetly not shipped (it is a module that crashes nsclient that I used a very very long time ago and should remove I guess...

lib_json is really a legacy windows "fix" as lib json cannot be built with old MSVC compilers.
So arguably should be "fixed" even on windows builds (except for legacy).

So yes sure, patch away, that would be very welcome...

As for the numbers this PR went a bit back and forth so they are likely not accurate...

PR #1546, opened 2026-09-18
Get Windows ARM64 building again, trim the RedHat matrix on PRs, add a Raspberry Pi package

mickem:
## Why the ARM64 build failed

The run that prompted #1223 is [24964804251](https://github.com/mickem/nscp/actions/runs/24964804251) (26 Apr, the last `main` build with ARM64 enabled). `build-server-arm64` failed at step 20, **Build Boost (static)** — and it could never have done anything else. `mickem/build-boost` shells out to `cmake-common`'s `boost-build --platform arm64`, and that project's `Platform` enum is exactly `x86 | x64 | auto`, so argparse rejects the value before a single file is compiled. (Logs are past the 90-day retention; the step name and the enum are the evidence.)

Boost was only the first wall. Everything below it that had to pick a target architecture was written as "x64, or else the 32-bit one", and none of it had ever been reached:

| Where | Was | Now |
|---|---|---|
| `actions/openssl` | `VC-WIN32` for anything not x64 | `VC-WIN64-ARM` |
| `actions/protobuf` | `-A Win32` | `-A ARM64`, plus a second **host-native protoc** — a cross-compiled one cannot run on the x64 runner |
| `actions/google-test` | `-A Win32` | `-A ARM64` |
| `actions/cryptopp` | `x64` for anything not x86 | `ARM64`; Crypto++ 8.9.0 also declares no ARM64 configuration at all, so `msbuild /p:Platform=ARM64` is rejected outright |
| architecture detection | five copies, four on `CMAKE_CL_64` ("64-bit pointers", so ARM64 reads as x64) | one `nscp_target_arch()` on `MSVC_<lang>_ARCHITECTURE_ID`, which is set under Ninja too |
| `wix.cmake` / installer | `-arch x64` and the x64 CRT merge module | `-arch arm64`; see the CRT note below |

Crypto++'s sources are fine on ARM64 — `config_asm.h` keys off `_M_ARM64`, and the hand-written assembly hangs off `CustomBuild` items already conditioned on `Win32`/`x64` — which makes the missing project configurations pure boilerplate. `build/python/msdev-add-arm64.py` fills them in, preserving CRLF and BOM so it stays a small diff. It also adds `neon_simd.cpp`, the one source the project omits that ARM64 needs: `cpu.cpp` calls `CPU_ProbeNEON()`, which nothing else defines, and on x86/x64 the call is never live.

Boost itself is built by a new in-repo action driving `b2` directly, mirroring `cmake-common`'s invocation — `--layout=system` and `stage/<platform>/<configuration>` above all — so the tree is interchangeable with what x86 and x64 still get from `mickem/build-boost`. `b2` is bootstrapped under a plain x64 environment, because the job's ambient one targets ARM64 and would produce a `b2.exe` the runner cannot execute; the build then runs under the *target* environment, or the DLLs fail to link and the static libraries silently archive x64 objects.

**Embedded Python is the one thing that does not cross-compile**: there is no ARM64 CPython that runs on an x64 host to link against. `NSCP_WITH_PYTHON=OFF` asks for the interpreter alone — the build still needs it for code generation — and `PythonScript`, Boost.Python and the bundled runtime drop out exactly as they do on a Linux host without libpython. The MSI names those files explicitly, so the same condition reaches `Product.wxs` as `-dHasPython`. This is a real difference from the x64 package and is called out in the README and an upgrade note.

**The ARM64 MSI does not bundle the Visual C++ runtime.** The runner has no `Microsoft_VC143_CRT_arm64.msm` — CI confirmed it, and the build now prints every `.msm` it can see when it comes up empty. A missing merge module is a hard error everywhere else; ARM64 opts out explicitly via `NSCP_ALLOW_MISSING_CRT_MERGE_MODULE`, and operators are told to install the ARM64 redistributable. Closing that gap properly means adding an ARM64 redist component to the build image, which is not in this PR.

Also fixes a typo that predates all of this: the artifact upload referenced `${{ inputs.platform }}`, which does not exist, so **no Windows build has been attaching its symbols zip** — while `release.yml` has been listing exactly that file.

## Only the newest RedHat on a pull request

Four Rocky jobs run on every PR and the two for Rocky 9 have almost nothing to say that the Rocky 10 pair does not — they differ in packaged library versions and in the python the integration tests run under (which is why the integration suite already skips Rocky 9), not in anything a change to NSClient++ breaks on its own. Release builds still produce all of them.

`include:` entries cannot carry an `if:` and a self-skipping matrix job still queues a runner, so the matrix is computed in a small job ahead of the build, selected by a `targets` input (`default` / `all`) — the same name and values `build-debian.yml` takes, so a value meant for one cannot be silently ignored by the other. "Newest" is read off the list rather than named.

The integration matrices are derived from the same lists rather than written out a second time. Both hand-written copies had already drifted: debian built the new Raspberry Pi entry and never tested it, and redhat would have failed at artifact download the first time a release was added.

## A Raspberry Pi package

Releases gain `NSCP-<version>-debian-trixie-arm64.deb`, built in a `debian:trixie` container on the arm64 runner. Trixie is Debian 13, the base of current Raspberry Pi OS 64-bit, so it installs on a Pi 3 or newer as well as on Debian 13 arm64 generally.

No 32-bit package, per the discussion on this branch: Raspberry Pi OS has defaulted to 64-bit since Bookworm, and neither route to armhf is worth it — an emulated build of ~820 translation units runs for hours against a 6h job limit, and a cross-compiled Debian armhf package still leaves out the ARMv6 boards (Pi 1, Pi Zero W) that Raspbian's own port exists for.

Debian does not package the .NET SDK, so the matrix carries a `dotnet` flag: the SDK is installed only where the distribution has it, the integration container skips the runtime there too, and the managed plugin API and its C# sample skip themselves — which they already knew how to do.

## Verification

The Windows ARM64 job ran green end to end on this branch: Boost, Crypto++, ~800 translation units, the link, the unit tests, signing, the MSI, cpack and the artifact upload. The Raspberry Pi package built and passed its integration tests on the same run. Both took several rounds of real CI to get there, and each failure is its own commit with the error it fixes quoted in the message.

Everything checkable without CI was checked locally: every YAML parses, both matrix scripts were run for each input value including the rejected one, `nscp_target_arch()` has a nine-case test across generators, the Crypto++ patcher was run against a real `cryptopp890.zip` with its output re-parsed as XML, the `check_nsclient` binaries were verified to be the architecture they claim by their PE machine type, and `NSCP_WITH_PYTHON` was exercised both ways through a real `cmake` configure.

The four "unknowns for CI" this PR opened with are answered: `check_nsclient` now publishes a `windows-arm64` asset (1.1.1, pinned by digest), `b2` does pick the ARM64 cross-tools once it runs under the target environment, WiX does accept `-arch arm64` given `InstallerVersion >= 500`, and `debian:trixie` does carry every dev package the build wants.

The temporary commit that ran these on pull requests (9f338106) has been **reverted** in 1ac75712 — reverted rather than dropped, since it sits several commits down and a squash merge would have carried it in. `build-feature.yml` is back to its original state apart from the `distros` → `targets` rename, and `build-main.yml` still carries ARM64 and `targets: all`.

Rebased onto `main` at 6d002dc1 to pick up the OpenSSL-as-DLLs and shared-plugin-code work. Two conflicts, both in `build-windows.yml`: main pinned `mickem/build-boost` by commit SHA where this branch had added an `arm64` guard (kept both), and main gave `installer_lib` its own static OpenSSL build where this branch had added the SDK pin (kept both). The OpenSSL DLL names main now derives are found by glob rather than by architecture, so `libcrypto-3-arm64.dll` needs nothing extra.

https://claude.ai/code/session_01RnWMsHBBNzvSNUXrVJVu7Y


mickem:
`build-redhat / integration-tests (rocky, 10, arm64)` is red on `0a7c9e13`, and I do not believe it belongs to this PR.

One assertion out of 1204 failed — [`plugin-threading.test.ts:447`](https://github.com/mickem/nscp/blob/main/tests/plugin-threading.test.ts#L447):

```
expect(servedWhileBlocked).toBeGreaterThan(0)
    Expected: > 0
    Received:   0
Tests: 1 failed, 362 skipped, 841 passed, 1204 total
```

Three things point away from this branch:

- **The test is timing-sensitive and has flaked on this exact runner before.** Its own comment says so: *"a slow runner can spend most of SLOW_MS on three NRPE round trips — which is exactly how this failed on the arm64 runner, at `2046 < 2046`"*. It was hardened for that in c54f1eb8 (15 Sep). `servedWhileBlocked == 0` means the first NRPE round trip did not complete before the slow check left the agent — the same class of timing, one notch further.
- **This branch changes no C++ and no tests.** The diff is workflow/action YAML for Windows ARM64, the RedHat/Debian matrix computation, CMake/WiX changes that are Windows-only or no-ops on Linux (`NSCP_WITH_PYTHON` defaults ON and configures identically; `FindCryptoPP` keys on `CMAKE_VS_PLATFORM_NAME`, empty on Linux), plus docs.
- **The same job was green earlier in this PR**, on identical Linux-path code (run [35329081807](https://github.com/mickem/nscp/actions/runs/35329081807)).

Runners have been heavily contended today — this PR's own run sat queued 45 minutes behind several others — which fits.

I have not re-run it yet on purpose: `build-server-arm64` is still building on this head, and if it fails I will push a fix, which re-runs this job for free. If ARM64 comes back green and this is the only red check, I will re-run it once then.

---

mickem:
`build-server-x64 / build` failed on `67ba903c`, in **Run tests**, on one assertion out of 214:

```
D:\a\nscp\nscp\modules\CheckDisk\check_disk_io_test.cpp(421): error: Value of: any_activity
  Actual: false
Expected: true
[  FAILED  ] DiskIoDataTest.LatencyIsComputedBetweenFetches (1647 ms)

99% tests passed, 1 tests failed out of 106
The following tests FAILED:
	 54 - check_disk_test (Failed)
```

**This is not this PR's.** The last x64 build before this push (`0a7c9e13`) was green, and the diff between the two heads is six files — `CMakeLists.txt`, `build/cmake/wix.cmake`, `installers/installer-NSCP/{CMakeLists.txt,Product.wxs,RUNTIME_DLL.wxs,setup.wxs}` — all packaging. There is no C++ in it, and the only top-level `CMakeLists.txt` hunk changes `VERSION_ARCH`, which feeds nothing but the MSI and ZIP file names. The translation units behind `check_disk_test` are byte-identical to the ones that passed.

**Why it flakes.** The test writes 16 MB, sleeps 1.1 s, re-samples the PerfDisk counters, and then requires at least one disk to show movement:

```cpp
boost::this_thread::sleep(boost::posix_time::milliseconds(1100));
data.fetch();
...
if (d.total_latency > 0.0) any_activity = true;
...
EXPECT_TRUE(any_activity);
```

Every numeric bound in that loop passed — `>= 0.0` and `< 60000.0` on all three latencies. Only the "something actually moved" check failed, which is the one part that depends on the host rather than on the code: `ofstream::flush()` hands the bytes to the OS, not to the platter, so on a runner with a write-back cache 16 MB can be absorbed without a counter-visible flush inside the window. Nothing in the assertion is wrong about the code under test; the stimulus just is not guaranteed to arrive in time.

**No fix exists to port**, so rather than widening a packaging PR into `CheckDisk`, here is the patch I would propose — retry the stimulus instead of assuming one round is enough, which keeps the assertion's meaning intact:

```cpp
  // A single 16MB write can sit in the write-back cache and never reach the
  // counters inside one interval, so give the disk several chances before
  // concluding that latency is not being computed at all.
  bool any_activity = false;
  for (int attempt = 0; attempt < 5 && !any_activity; ++attempt) {
    write_block(path);                     // the existing 16MB write
    boost::this_thread::sleep(boost::posix_time::milliseconds(1100));
    data.fetch();
    for (const auto &d : data.get()) {
      EXPECT_GE(d.read_latency, 0.0);
      EXPECT_GE(d.write_latency, 0.0);
      EXPECT_GE(d.total_latency, 0.0);
      EXPECT_LT(d.read_latency, 60000.0);
      EXPECT_LT(d.write_latency, 60000.0);
      EXPECT_LT(d.total_latency, 60000.0);
      if (d.total_latency > 0.0) any_activity = true;
    }
  }
  std::remove(path.c_str());
  EXPECT_TRUE(any_activity);
```

Opening the file with `FILE_FLAG_WRITE_THROUGH` would work too and is arguably more direct, but it needs a raw `CreateFile` handle in place of the `ofstream`.

I am holding the one re-run until `build-server-arm64` reports, which is the job this PR exists to prove — if that one needs another fix, the push re-runs x64 for free and the re-run stays unspent. Otherwise I will re-run this job on its own.

---

PR #1548, opened 2026-09-18
Close every Low finding from the 0.20.0 security audit

mickem:
The 0.20.0 security audit left 26 Low findings open or partly fixed. This closes them, grouped by area below with the user-visible behaviour changes called out.

## Outbound clients

The host-override guard learned two more classes of key. **Transport security** (`verify mode`, `insecure`, `insecure-skip-verify`, `use psk`, `security`, `ssl`, `no ssl`, `tls version`, `ca`, `certificate`, `certificate key`, `certificate format`, `allowed ciphers`, `dh`) decides whether the host receiving a configured credential has to prove it is the configured one, and was a request option on most client modules. **Message addressing** (`recipient`, `sender`) let a caller send mail of their choosing through the target's authenticated mailbox; it gets its own narrower opt-in, `allow recipient override`.

A failed CA load no longer reports the path or the OpenSSL reason to the caller, which answered "does this file exist and can the service read it?" for any path with the agent's privileges. The exception types carry a log-only detail beside the caller-safe message.

The NSCA and NRPE clients clamp a request-supplied payload length to the protocol maximum. `payload-length=2147483647` used to allocate about 2 GB per payload.

## Script execution and check arguments

A NUL in an argument is refused whatever `allow nasty characters` says: `CreateProcessW` reads the command line as a C string, so a NUL dropped every operator-fixed argument after the substitution point. `ext-scr add --import` is confined to the script root, `${shared-path}` and the upload staging area, so the show/delete sandbox cannot be walked around by carrying a file inside it first. The Windows read loop takes only what `PeekNamedPipe` reported, so a script printing in exact buffer-sized chunks cannot park a worker past the timeout, and both process handles are owned by RAII rather than leaked on the timeout path. `check_docker host=` must equal the configured endpoint. `check_nt` FILEAGE passes `max-depth=0` instead of listing a whole tree.

## Listeners and TLS

The insecure NRPE cipher string says what the mode is, instead of carrying an `!ADH` that never excluded AECDH and could not have excluded anything without breaking a mode that loads no certificate; the secure preset gains `!aNULL`. `nscp nrpe install` writes `use ssl`, the key the server actually reads. `allowed hosts` implements the `*` ranges it has always documented, and an unparseable numeric entry is an error rather than an exception that stopped the listener from starting.

## Web server and UI

Every response carries `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy` and a content security policy, plus HSTS over TLS. The beast backend (every Linux package) can negotiate TLS 1.3 and honours the `tls version` and `allowed ciphers` settings the other listeners use. A session token is minted only by the login routes, so a monitoring poll no longer evicts live UI sessions from the 4096-entry store. The session identity moved from the response cookie map to request-scoped context, so the bearer is no longer echoed back as a `Set-Cookie` nothing read, and the UI keeps it in `sessionStorage`. Handler exceptions log their detail and answer a bare 500 without leaking the response object. `/api/v2/scripts/<runtime>` is an allow-list, and `lua` works for the first time. The web bundle builds with `npm ci` behind an `npm audit` gate.

## Core and packaging

A module is named by a single file name: an absolute or traversing `[/modules]` entry no longer loads an arbitrary shared object, the `./modules` fallback resolves against `${exe-path}`, and Windows loads modules under `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS`. A settings `Control.LOAD`/`SAVE` cannot name an `http://` store. Redaction matches credential-shaped key names as well as what the loaded modules registered, and the NRDP `key` alias is registered as a password. The systemd unit is sandboxed, the DEB drops its `sudo` dependency, and CauseCrashes is behind `-DBUILD_TESTING_MODULES`.

## Fleet and modules

The certificate pin is enforced as a pin: a leaf is matched by its SPKI digest, and a pinned PEM that is itself a CA keeps hostname verification instead of becoming "trust this CA for any name". A renewal may only rotate the bundle signing key when the old key endorses the new one. The fleet guide states plainly that a fleet server can run code as `SYSTEM` or `root` on every enrolled host, and the checks that connect elsewhere are documented as remote-connection primitives.

## Testing

New unit tests cover the client guards, the payload clamp, wildcard hosts, pinned certificates, renewal rotation, the docker endpoint, sensitive key names and session tokens, plus two new integration suites. The full ctest suite and 286 integration tests pass.

Six security notices and twelve upgrade notes describe the behaviour changes.

## Notes for reviewers

`NoNewPrivileges` is deliberately left off the systemd unit and documented, since the launcher tells operators to use `sudo -n -u` to drop privileges for a script.

Three behaviour changes need operator action and carry upgrade notes: the transport and recipient guards refuse request options that used to be honoured, `[/modules]` entries naming a path stop loading, and CauseCrashes is no longer in release packages.


mickem:
Closing in favour of a four-way split — 109 files in one review was too much.

| | Part | Findings | Files |
|---|---|---|---|
| 2 | Script arguments, check targets, listener TLS — #1551 | X-4 – X-8, S-2 – S-5, M-7 | 27 |
| 3 | Web server browser surface and sessions — #1552 | W-3 – W-6, U-3 – U-5 | 36 |

The parts are stacked, each based on the one before, so every diff shows only its own work. They are stacked rather than independent because the TLS helper and the HTTP client each carry changes from more than one group; four branches off `main` would have conflicted with each other on merge. Merging in order is conflict-free.

Two documents needed splitting by hand, since the notices did not map one-to-one onto the parts. The fleet notice covered the certificate pin, the signing key and the trust boundary, so it became a pinning notice in part 1 and a trust-boundary notice in part 4, with matching upgrade notes. The remote-connection-checks section moved into the script notice in part 2.

Nothing was dropped or changed in the split. Every one of the 110 changed files is assigned to exactly one part, and reassembling all four onto current `main` reproduces the tree this pull request carried, byte for byte, apart from the one notice that was retitled. The stack was also rebased onto current `main`, 17 commits further along than this branch.

Each part was compiled and tested as it was stacked: four clean builds, all unit tests passing at every step, plus the web unit, end-to-end and audit checks and the documentation notes validator.

---

mickem:
Correction to the comment above: the four parts are **independent**, not stacked, and all four target `main`.

I stacked them initially because the TLS helper and the HTTP client looked like they carried changes from more than one group. Grouping the fleet certificate pin with the client transport work — it lives in those same two files — removed the overlap, so every file belongs to exactly one part. Checking that through, no part uses a symbol another part introduces either: the one changed signature, `configure_ssl_context`, took a defaulted parameter, so the listener call sites are untouched.

That means each part can be reviewed and merged on its own, in any order. Cherry-picking each onto `main` was clean, and an octopus merge of all four reproduces the tree this pull request carried.

It also matters for CI: the full build matrix only runs for pull requests into `main` or `feature/**`, so as a stack, parts 2 to 4 were getting three checks instead of twenty and their Windows-only code would not have been compiled until the parts ahead of them merged. Independent, all four get the full matrix now.

---

PR #1551, opened 2026-09-18
Audit Low 2/4: confine script arguments, check targets and listener TLS

mickem:
Eight Low findings from the 0.20.0 audit, covering the script launcher, the checks that take a path or a connection as an argument, and the listener TLS presets — plus one correctness fix that the audit had filed as a finding and should not have.

## Script execution

**A NUL in an argument is refused** whatever `allow nasty characters` says. Neither metachar filter contained it, protobuf strings carry it and REST forwards parameters verbatim, so `CreateProcessW` — which reads `lpCommandLine` as a C string — silently dropped every operator-fixed argument after the substitution point. A template of `check.exe $ARG1$ --read-only` ran without `--read-only`.

**`ext-scr add --import` is confined** to the script root, `${shared-path}` and — only for a file the service itself owns and no one else can write — the temp directory the REST upload route stages into. Notice 050 sandboxed `show` and `delete`; `--import` walked around that by copying the file in first. `${temp}` is shared, so the directory alone is not a boundary: a local account can drop a file there and match any name the staging route would pick. Ownership is what separates the two. Windows is admitted on the directory check alone and says so — reading the DACL belongs in a change that can be tested there.

**The Windows read loop takes only what `PeekNamedPipe` reported** instead of looping while the last read filled the buffer, so a script printing in exact 4086-byte chunks can no longer park a worker past the timeout. The final drain after the child exits has its own deadline, separate from the command timeout, so a process that never stops writing cannot hold the loop open past the kill — and one that exited normally is still owed its remaining output. `readFromFile` builds the chunk from the byte count rather than as a C string, so output is no longer cut at the first NUL (for UTF-16, the second byte). The process handle is owned by RAII rather than leaked on the timeout branch.

## Check targets

`check_docker host=` must name the configured endpoint; it let any query caller point the agent at any unix socket or named pipe and read back status codes and parse errors. The refusal happens before any connect or shape check and does not echo the value it refused.

**`check_nt` FILEAGE is a correctness fix, not a security one.** The audit filed it as M-7 and the first version of this branch wrote it up as an information disclosure. That is over-claimed and has been reclassified (`4c415d9`): check_nt is an authenticated protocol whose purpose is to answer questions about the host, `FILEAGE` takes a path by design, and a caller allowed to run it is not crossing a boundary by naming a directory. What is genuinely wrong is the number: `FILEAGE` answers with a single age taken from `perf(0)`, so the old mapping onto `check_files` — which walks a tree — reported whichever file the walk emitted first, neither the oldest nor the newest. The mapping now resolves to `check_single_file`, which stats exactly one path, so there is only ever one candidate and the age is defined. The code and tests are unchanged by the reclassification; only the upgrade note and the security notice moved.

## Listeners

**The insecure NRPE cipher string says what the mode is.** `!ADH` excluded only DH-anon while the anonymous elliptic-curve suites are AECDH, and since the mode loads no certificate the anonymous suites are the only ones that can complete a handshake — the exclusion could not have held without breaking the mode. The secure preset gains `!aNULL`, and `nscp nrpe install` recognises both spellings so a re-run no longer warns about a downgrade it does not perform.

**`nscp nrpe install` writes `use ssl`**, the key the server actually registers, rather than `ssl`, which nothing consumed. On a host where an operator had set `use ssl = false`, the command reported encryption and client-certificate authentication while the listener stayed plaintext.

**`allowed hosts` implements the `*` ranges its help text has always advertised**, and an unparseable numeric entry becomes an error rather than an exception escaping `refresh()`, which dropped the module and left the listener down.

## Documentation

`check_wmi`, `check_mysql`, `check_mssql` and `check_uncpath` are documented as the remote-connection primitives they are, in the securing guide and the check-access page. The FAQ no longer tells readers `--insecure` writes `ssl = true`.

## Testing

Covered by `execute_process_test`, `docker_endpoint_test`, `check_nt_commands_test`, `nrpe_module_test` and `allowed_hosts_test`, plus the docker cases in `tests/checkdocker-commands.test.ts` run against the built binary. Note that the Windows launcher changes are compiled only by the Windows jobs here; the local verification was Linux.

---

**Part 2 of a four-way split of #1548**, which closed all 26 open and partly-fixed Low findings in one 109-file change.

| | Part | Findings | Files |
|---|---|---|---|
| 2 | Script arguments, check targets, listener TLS (this one) | X-4 – X-8, S-2 – S-5, M-7 (reclassified) | 29 |
| 3 | Web server browser surface and sessions — #1552 | W-3 – W-6, U-3 – U-5 | 36 |

**The four are independent.** They touch disjoint sets of files, and no part uses a symbol another introduces, so each can be reviewed and merged on its own, in any order. An octopus merge of all four onto `main` is clean and reproduces the tree #1548 carried. Each part carries its own security notices, upgrade notes and tests, and each validates on its own.


mickem:
`build-server-x64 / integration-tests` failed on `0cb5736`, and it is not this PR's either — though it took ruling out, because this PR does touch the Windows process runner.

```
● nscp test console › keeps running commands after a module reload
  expect(r.exitCode).toBe(0)
  Expected: 0
  Received: 3221225477
```

`3221225477` is `0xC0000005`, an access violation. The console ran the command after the reload (`expect(out).toContain("after-reload")` passed, and it did not time out) and then died on the way out.

**None of this PR's code is loaded when that test runs.** `tests/src/nscp.ts:296` applies settings cumulatively — each `configure` writes individual keys into the same ini — so by the time the reload case runs, the fixture's `/modules` section reads:

| Module | State | Touched by this PR |
|---|---|---|
| CheckHelpers, CheckDisk | enabled | no |
| CheckExternalScripts | **disabled** (`command-client-console.test.ts:305`) | yes |
| NRPEServer, NSCAServer, NSClientServer, WEBServer | **disabled** (same line) | yes (NRPEServer, NSClientServer) |
| CheckDocker, NRPEClient | never enabled | yes |

That covers every module in the diff. The two changes outside `modules/` are confined the same way:

- `include/process/execute_process_w32.cpp` — `grep -rl execute_process` outside `include/process/` names exactly one consumer, `modules/CheckExternalScripts`, which is disabled here.
- `include/net/socket/allowed_hosts.cpp` — `refresh()` is called by the listener modules, all of which are disabled here.

**What the crash is actually in.** The test is itself a regression test for a crash in `CommandClient::loadModuleEx`, as its own comment records: a reload used to replace the client object the console was built on and leave the completion hooks holding a freed pointer. That path — a queued `delayed,service` reload landing while the console is dispatching a command — is racy by construction, and it is untouched by this PR. On `main` at this PR's merge-base (`17ab090`, run 35386761868) the same `build-server-x64 / integration-tests` job is also red, on a different timing assertion (`plugin-threading › keeps a different module answering while one module is blocked`, `servedWhileBlocked > 0`), so that job is not currently green on the base branch either.

I have not widened this PR with a fix for a crash in code it does not touch. Re-running the failed job is blocked while the rest of the run is in flight (`403 This workflow is already running`); I will spend the one re-run as soon as it clears, and keep watching this PR until it is green.

---

mickem:
`build-server-x64 / integration-tests` passed on `4c415d9`, including `nscp test console › keeps running commands after a module reload`. The access violation did not recur on a head that still carries every one of the changes it was reported against — the only difference between the two heads is the FILEAGE documentation reclassification, which touches no code. That matches the diagnosis above: the crash is in the `CommandClient` reload race, in a module set this PR's fixture does not load.

All 28 checks are green and the PR is mergeable. Nothing outstanding from my side — it is waiting on review.

---

PR #1552, opened 2026-09-18
Audit Low 3/4: harden the web server's browser surface and session handling

mickem:
Seven Low findings from the 0.20.0 audit, covering the REST server, the shipped UI and the way the web bundle is built.

## Browser hardening headers

Every response now carries `X-Frame-Options`, `X-Content-Type-Options` and `Referrer-Policy`, a content security policy, and `Strict-Transport-Security` when TLS is on. Responses previously carried only `Content-Type` and cookies, so a page that could frame the agent got an authenticated UI to click-jack.

## TLS 1.3 on Linux builds

The beast backend — every Linux package — asked asio for `tlsv12_server`, which pins *both* ends of the range, so the REST API could never negotiate TLS 1.3 and ignored the `tls version` and `allowed ciphers` keys the NRPE and NSCA sockets honour. It now takes a minimum version and reads both keys.

## A session token is minted only by the login routes

`process_auth_header` and `process_password_header` created an eight-hour token on **every** request, so each `check_nscp_api` poll filled the 4096-entry store: a host running twenty checks a minute evicted the operator's UI session within hours, and any authenticated user could do it deliberately.

The session identity moved from the response cookie map to request-scoped context, which also stops the bearer being echoed back as a `Set-Cookie` that no request path ever read — the credential was stored twice, and the cookie looked like a working session mechanism to anyone reading the code next.

The UI keeps the bearer in `sessionStorage` rather than `localStorage`, so it no longer survives a browser restart or travels in a profile backup, and logout clears it eagerly. The Playwright fixture seeds the same place the app reads, and the mocked login route answers the logout revoke.

## Handler exceptions

The `what()` text — `make_address` on an unavailable peer, `bad_optional_access`, `regex_error` — went verbatim to an unauthenticated caller, and the response object was leaked on every throw. Exceptions now log their detail and answer a bare 500.

## The scripts runtime segment is an allow-list

The first URL segment was used verbatim both as the grant suffix and as the exec target, so a role granted `scripts.*` could drive the `add --import` and `delete` verbs of any loaded module implementing them, and a PUT or DELETE with an empty segment called `exec_command("")`. Mapping the segment also makes `lua` work for the first time: `get_runtimes` has always advertised it and this function never translated it.

## Build

The web bundle builds with `npm ci` behind an `npm audit` gate, so the bytes in the zip and in the MSI match `package-lock.json`.

## Testing

Covered by `session_manager_interface_test`, the AppNavbar and Login suites, and two new integration suites for the headers and the runtime allow-list. Locally: C++ suite green, 122 web unit tests, 18 Playwright end-to-end tests, `tsc` clean, `npm ci` and `npm audit` clean.

---

**Part 3 of a four-way split of #1548**, which closed all 26 open and partly-fixed Low findings in one 109-file change.

| | Part | Findings | Files |
|---|---|---|---|
| 2 | Script arguments, check targets, listener TLS — #1551 | X-4 – X-8, S-2 – S-5, M-7 | 27 |
| 3 | Web server browser surface and sessions (this one) | W-3 – W-6, U-3 – U-5 | 36 |

**The four are independent.** They touch disjoint sets of files, and no part uses a symbol another introduces, so each can be reviewed and merged on its own, in any order. An octopus merge of all four onto `main` is clean and reproduces the tree #1548 carried. Each part carries its own security notices, upgrade notes and tests, and each validates on its own.


mickem:
`build-server-x64 / integration-tests` failed on `87e1632`. I cannot say yet whether it is this PR's, and I am not going to guess — here is exactly what is known and what I did.

**There is no failure to read.** GitHub returns `HTTP 404` for the job's log archive (three attempts, ~20 minutes apart) and the check run carries an empty `output`. What the job record does show is unusual:

| Step | State |
|---|---|
| 1–13 (checkout … `Run REST scenarios (no-docker)`) | `success` |
| 14 `Run installer tests` | **`in_progress`** |
| 26–28 (`Post` steps) | `pending` |

The job is `completed` / `failure` with **no failed step**. It started 22:04:20 and ended 23:15:23, with step 14 alone running from 22:28:32 — about 47 minutes. That is the shape of a killed or lost runner, not a test assertion.

**Two candidate causes, and I can't separate them without the log:**

1. *Environmental.* `tests/msi/helpers.py:10` wraps every command in `subprocess.run(..., shell=True, capture_output=True, timeout=120)`. On Windows that timeout does not do what it looks like: it kills the `cmd.exe` child, then `communicate()` blocks waiting for EOF on pipes the detached `msiexec` grandchild still holds. So one hung `msiexec` hangs the whole step forever regardless of the 120 s bound. Two of the MSI cases fetch their configuration over the network mid-install (`CONFIGURATION_TYPE=https://raw.githubusercontent.com/...`), which is a plausible thing to hang on from a runner.
2. *This PR.* It touches the web server, so a service that will not stop during the uninstall could wedge `msiexec` the same way.

Against (2): the preceding step, `Run REST scenarios (no-docker)`, passed — that one actually starts the agent and drives the REST API, so the web server built, served and shut down fine. The diff to `ServerBeastImpl` adds TLS option parsing and headers and changes nothing in the start/stop path (`grep` for `stop|join|thread|close|shutdown` across it returns one line, a `if (thread_)` guard in the new `setTlsOptions`). And `build-server-x64 / integration-tests` reached and passed this same step on #1553.

So the evidence leans environmental, but "leans" is not "is", and the failing step is one this PR could in principle affect. I have re-run the failed job — that is this PR's one re-run, spent to regenerate a log I can actually read as much as to retry it. If it hangs again at `Run installer tests`, that is reproducible and I will treat it as this PR's and dig into it rather than re-run again.

---

mickem:
The re-run passed, and the timings settle it: `Run installer tests` took **4 m 14 s** (23:43:35 → 23:47:49) on attempt 2, against the ~47 minutes it was stuck for on attempt 1 before the runner was killed. Same commit `87e1632`, same step, an order of magnitude apart — the step wedged rather than failed, which rules out this PR's diff as the cause. CI is green across all 28 checks and the PR is mergeable.

One thing worth leaving behind, since it turned a single stuck `msiexec` into a dead job and an unreadable log: `run_with_timeout` in `tests/msi/helpers.py:10` cannot actually time out on Windows. `subprocess.run(..., shell=True, capture_output=True, timeout=120)` kills the `cmd.exe` it spawned, then blocks in `communicate()` waiting for EOF on pipes the detached `msiexec` grandchild still holds open — so the 120 s bound is never enforced and the step hangs until the job is killed. Spawning without `shell=True`, or reading the pipes with an explicit deadline and killing the whole process group on expiry, would turn a hang like this into a 2-minute failure with a log attached instead of 47 minutes and nothing.

---

mickem:
`build-server-x86 / build` failed on `00692d1`. It is not a compile error, and the rest of the head answers the question the previous failure raised.

**The C1083 is fixed, confirmed by CI.** `build-x86-static / build` — the job that failed on `e2ecf5f` with `Cannot open include file: 'openssl/ssl.h'` from `include/net/tls_versions.hpp` — **passed** on this head. So did `build-server-x64 / build`, which compiles the same sources with the same MSVC. The include now sits inside the `USE_SSL` guard, where `SyslogClient` and `CollectdClient` (which compile `socket_helpers.cpp` privately *without* SSL in the static build, by design) never reach it.

**This failure is the killed-runner shape, not a build error:**

| Step | State |
|---|---|
| 1–29 (checkout … `CMake (NSCP)`) | `success` |
| `Build nsclient` | **`in_progress`** |
| `Run tests` and everything after | `pending` |

No failed step, and the log archive stops at 03:54:52 while the job was marked failed at 04:41 — the upload was cut off mid-build. That is the same signature as the `Run installer tests` job on this PR's previous head, which passed on a re-run in 4 minutes after hanging 47. `build-server-x86 / integration-tests` is `skipped` as a consequence, not on its own merits.

The evidence that it isn't the diff: the identical source compiled clean on `build-server-x64` (same toolchain) and on `build-x86-static` (same architecture, and the stricter configuration that actually caught my real bug last round).

Re-running is currently refused (`403 This workflow is already running`) — the x64 and x86-static integration suites are still going. This failure qualifies for a re-run under "died before any test body ran"; I will spend it as soon as the run clears, and keep watching until the PR is green.

---

PR #1561, opened 2026-09-19
build: link the Boost.JSON library instead of compiling it in

Fantu:
Follow-up to #1544, from your reply there: `libs/json` goes away and the tree links the Boost.JSON library that every build already has.

## The library is already there

`json` has been a required component of `find_package(Boost)` since the json_spirit migration (#971, May 2025), which is also the commit that added `include/json/use_json.cpp`. So the build has been demanding the library and compiling its own copy of it side by side ever since; #1544 only narrowed that to a single copy in `nscp_json`.

If the component were missing, `Boost_FOUND` would be false and configure would fail, so every green job already has it:

- Linux and RedHat: the distribution package.
- Windows: `mickem/build-boost` is asked for `json` in `build-windows.yml`.
- Windows legacy XP: the same, in `build-windows-legacy.yml`, with `toolset: msvc141`.

That last one is why this does not keep the legacy exception you suggested: the XP job has been building Boost.JSON with the old toolset all along. I cannot build MSVC here, so if that reading is wrong the XP check is what will say so, and the exception is easy to add back.

## Changes

- `nscp_json` dropped; the 26 places that listed it now list `${Boost_JSON_LIBRARY}`, the same form the tree uses for the other Boost libraries.
- `BOOST_JSON_DYN_LINK` added to the dynamic-runtime global defines, beside the thread and program-options ones, so `BOOST_JSON_DECL` resolves to `dllimport`. The static build leaves it undefined and links Boost.JSON statically, as it does the rest of Boost.
- The MSI ships `boost_json.dll` in place of `nscp_json.dll`, and `JSON` joins the Boost DLL copy list in the top-level `CMakeLists.txt`. The MSI install test and the unreleased upgrade note follow.
- `include/json/use_json.cpp` stays, for `installer_lib` alone: its custom actions run inside msiexec before anything is installed and link a static Boost built with `system` and `filesystem` only, so there Boost.JSON stays header-only. There is a comment on the entry saying so.
- `build.md`: the legacy instructions build their static Boost without `--with-json`, although the legacy CI job has always asked for it. Added, with `--with-container` to match the shared build above.

## Verification

Built in a Debian sid chroot, agent-only configuration (WEBServer and DotnetPlugins off, 35 modules), against the same configuration on the merge base. Stripped binaries, no debug symbols in the numbers:

| | base | this branch |
|---|---|---|
| `nscp` + 35 modules + libraries | 52,120 KB | 51,680 KB |
| the 35 modules alone | 46,272 KB | 46,272 KB |
| `libnscp_json.so` | 437 KB | gone |

The whole difference is `libnscp_json.so` disappearing; every module is byte-for-byte the same size. That is the point rather than a disappointment: after #1544 there was only one copy left, and what this removes is a private copy of a library the system already ships (`libboost_json.so.1.90.0`, 414 KB on sid).

- 106/106 unit tests pass. `settings_query_handler_test`, `plugins_test` and `icinga_client_test` link `libboost_json` and are among them.
- No new unresolved symbol in `nscp`, the libraries or the modules: `ldd -r` reports the same four Boost.Thread symbols in `CheckExternalScripts` on both sides, and nothing else anywhere.
- `python3 docs/hooks/notes.py --check` passes.

Windows is untested locally, so the Windows, XP and MSI checks are the real test of the `dllimport` define and of the installer file list.


mickem:
Hmm... 

I am pretty sure it is required on windows as we are targeting an old SDK there and it is not supported (or it has dependencies on something that is not supported on legacy windows). 
So the library is needed and used on Windows but we can put a gate in place and on Linux (and perhaps modern windows) use boost one:

I.e. 
```cmake
# example
if(win32)
set (json_lib nscp_json)
else()
set (json_lib ${Boost_JSON_LIBRARY})
endif()
```

mickem:
interesting... build succeeded...

I did not expect that...

Fantu:
The legacy job is the one I would have bet on too, so it is worth saying exactly what went green, since it covers more than compilation.

All three Windows jobs run `integration-tests-windows.yml`, which runs the `tests/msi` suite through the `installer-test` action. That suite installs the built MSI and checks the installed file list, and `tests/msi/tests/normal-install.yaml` now lists `boost_json.dll`. So `build-x86-static`, the XP job with `msvc141` and static Boost, built the tree, produced the MSI, installed it and found the library where the installer says it is. The x64 and x86 jobs did the same for the dynamic runtime, which is what exercises `BOOST_JSON_DYN_LINK` and the `dllimport` path.

The reason it worked is in `build-windows-legacy.yml:170`: `json` has been in the list of Boost libraries that job asks `mickem/build-boost` for since the job was written, so that toolchain has been building Boost.JSON all along. Nothing here asks the old SDK for anything new; the branch only stops compiling a second copy of a library the build already produced.

So I would leave the gate out. It would mean keeping `nscp_json` alive and maintaining two configurations, and the difference it exists to protect is the one the checks now cover on every push. If a future toolchain does break on it, the gate is a few lines and comes back easily.

mickem:
I'm gonna fire up a Vm and verify that it actually works on old windows in the next few days...

PR #1568, opened 2026-09-20
feature: verify TLS peer certificates by default in client modules

mickem:
## Summary

This change makes TLS peer certificate verification the default for three outbound client modules (`NSCPClient`, `CheckMKClient`, and `check_nsclient_web_online`) that previously accepted any certificate presented by the remote server. Since these clients send credentials (passwords or API tokens) over TLS, an unauthenticated connection exposes those credentials to on-path attackers.

## Key Changes

- **NSCPClient**: Changed `verify mode` default from `none` to `peer`, and `ca` default to `${ca-path}` (the agent's trust bundle). The target carries the remote agent's REST API password, so verification is now required by default.

- **CheckMKClient**: Added `verify mode = peer` and `ca = ${ca-path}` defaults when TLS is enabled. Previously `verify mode` had no default (empty string parsed as `verify_none`).

- **check_nsclient_web_online** (CheckNet): Changed `verify` default from `none` to `peer` with `ca` defaulting to `${ca-path}`. This check sends the remote agent's password in headers.

- **Settings infrastructure**: Extended `nscapi::targets::target_object::add_ssl_keys()` to accept an `ssl_defaults` struct, allowing modules to register secure defaults for `verify_mode` and `ca` with the settings layer. This ensures defaults are applied even when nothing is configured, and allows `${ca-path}` macros to be expanded by the settings layer.

- **Blank value handling**: Added logic to treat blank `verify mode` and `ca` values as "not set" rather than silently disabling verification. An empty string now falls back to the default; `verify mode = none` is the explicit opt-out.

- **Path expansion**: Ensured `${ca-path}` macros in defaults and command-line arguments are expanded through the handler's `expand_path()` method, since the settings layer only expands configured values.

## Implementation Details

- The `ssl_defaults` struct in `nscapi_targets.hpp` allows modules to declare secure defaults that are registered with the settings layer, ensuring they apply to unconfigured targets and that path macros are expanded.

- Connection data constructors now handle blank values explicitly: if `verify_mode` or `ca_path` is empty after reading from configuration, they fall back to the defaults.

- Tests verify that:
  - Default targets verify the peer against `${ca-path}`
  - Explicit `verify mode = none` or custom `ca` values are honored
  - Blank values fall back to defaults rather than disabling verification
  - Untrusted certificates are rejected by default
  - `verify=none` still allows connection to self-signed certificates

## Migration Notes

Existing deployments pointing at agents with self-signed certificates will now fail the handshake. Users must either:
1. Point `ca` at the self-signed certificate and use `verify mode = peer-cert`
2. Point `ca` at their own CA and keep `verify mode = peer`
3. Explicitly set `verify mode = none` to restore the previous unverified behavior

See the upgrade documentation for configuration examples.

https://claude.ai/code/session_01NeicrerwKQDEjLjoWkLtMu


mickem (commented):
## Review

The change is sound and the mechanism is the right one. I traced the whole chain and it holds up:

- `typed_string_value::notify()` (`include/nscapi/settings/helper.cpp:64-77`) seeds the read with `default_value_` when `has_default_` is set and then runs the `post_functor`, so a `path_fun_key` default really does get `expand_path()`'d — the `ssl_defaults` plumbing is not just decoration.
- The blank-value handling is a genuine hole being closed: `get_verify_mode()` maps `""` to `verify_none`, so `verify mode =` in an ini was a silent opt-out.
- Peer verification on both transports does pull in hostname checking (`include/net/socket/client.hpp:310-317` for the socket path, `apply_verify_callback()` for the HTTP path), so `verify mode = peer` is a real check, not just chain validation.
- Missing-CA behaviour differs by transport but neither breaks the opt-out: `configure_ssl_context()` only *records* a load failure into `errors` (`include/net/socket/socket_helpers.cpp:549-552`), and `set_verify_mode()` has already run, so `verify mode = none` still connects. `${ca-path}` is always a file (never a `-CApath` directory) per the cmake detection at `CMakeLists.txt:1164-1194`, so `load_verify_file()` is the right call.
- Notes validate (`python3 docs/hooks/notes.py --check` → OK), numbering and the `321`/`061` cross-links follow the `320-check-tcp-unverified-tls` / `060-check-tcp-verifies-by-default` precedent exactly, and the anchor slug matches (`check_mk` keeps its underscore through the toc slugifier).

I could not compile here (no boost/cmake in this container), so the C++ below is read, not built. CI is green on all 28 checks including both Rocky 10 integration jobs.

Five things worth acting on, roughly in order.

### 1. The documented opt-out gets no warning, unlike every other client that has one

`NRDPClient`, `IcingaClient`, `Op5Client`, `ElasticClient` and `NRPEClient` all log once per target when the operator turns verification off — `socket_helpers::client_verify_mode_disables_verification()` exists for exactly this, and `NRPEClient::warn_if_unverified()` (`modules/NRPEClient/nrpe_client.hpp:155-170`) is the fullest version, with the bounded once-per-target set.

This PR promotes `verify mode = none` from "the default" to "a deliberate choice", on two paths that carry a password, and adds no warning. That leaves `NSCPClient`, `CheckMKClient` and `check_nsclient_web_online` as the only credential-carrying clients that can be silently unverified. Since the point of the change is to bring these three in line with the others, the warning is the matching half.

### 2. The migration advice omits the hostname requirement, which is the failure operators will actually hit

`ssl_connection::connect()` installs `boost::asio::ssl::host_name_verification(host)` whenever `verify_peer` is set. So pointing `ca` at the agent's self-signed certificate is *necessary but not sufficient* — the certificate also has to carry a SAN matching the address the target is configured with.

`build_subject_alt_name()` (`include/net/socket/socket_helpers.cpp:723-753`) puts in the machine's host name plus **one** discovered IPv4 and **one** IPv6 address. A target pointed at that agent by any other address — a second NIC, a NAT'd or public address, a CNAME, a host name the agent doesn't know itself by — fails with a certificate-verify error even after `ca` is correct, and the upgrade note's worked example is exactly that shape:

```ini
address = nscp://192.168.56.103:8443
ca = /etc/nsclient/agent-web01.pem
verify mode = peer-cert
```

That works only if `192.168.56.103` is the address the remote agent discovered for itself. `docs/upgrades/next/061-…`, the `NSCPClient.md` reference and the `network-checks.md` scenario (which uses `url=https://192.168.0.10:8443`) all need a sentence saying the SAN has to match, and what to do when it doesn't (regenerate the certificate, or use the name the agent knows itself by). This is the most likely support question the change creates.

### 3. `verify mode = peer-cert` is presented as the fix for self-signed, and on a client it does nothing

`peer-cert` sets `verify_peer | verify_fail_if_no_peer_cert`, and OpenSSL documents `SSL_VERIFY_FAIL_IF_NO_PEER_CERT` as *ignored in client mode*. So for these outbound paths `peer-cert` is exactly `peer`; what makes a self-signed peer verify is `ca` pointing at it. The upgrade-note table reads as though `peer` would fail and `peer-cert` is required:

| The remote agent still uses its self-signed certificate | `ca = <that certificate>` and `verify mode = peer-cert` |

I know this idiom is already project-wide (`NRPEClient`'s warning text uses it too), so this is a clarity nit rather than a defect — but it appears in six new/edited doc passages here, and `add_ssl_keys()`'s new shared `verify mode` description now teaches it to every target in the generated reference. Either say plainly that `ca` is the part that matters, or drop the `-cert` suffix from the client-side advice.

### 4. Six default literals across three layers, per module

For `NSCPClient` the pair `"peer"` / `"${ca-path}"` is now written in `nscp_handler.hpp` (ctor `set_property_string`), again in `add_ssl_keys(root_path, {"peer", "${ca-path}"})`, and again in `nscp_client.hpp` (`get_string_data` defaults **and** the blank fallbacks). `CheckMKClient` repeats the same six. Changing the default means finding twelve string literals, and a mismatch between layer 2 and layer 3 would be invisible (the settings layer would apply one default, the CLI/REST path another).

Two `constexpr` in `nscapi_targets.hpp` next to `ssl_defaults` would collapse all of it. The blank-fallback-plus-expand block is also byte-for-byte identical in both `connection_data`s and would sit naturally alongside `client_verify_mode_disables_verification()`.

### 5. Smaller items

- **`check_mk_client.hpp:79`** — the comment says the handler "is the same handler the connection itself is driven through (see `check_mk_client_handler::query`)", but `send()` constructs a second `std::make_shared<client_handler>()` (line 123). Stateless, so harmless, but the comment asserts something false and the query builds two handlers where one would do. Build it once in `query()` and pass it to `send()`.
- **`check_mk_client.hpp:22`** — `handler = nullptr` as a *default argument* makes "no expansion" the path a new caller gets by accident, and the failure mode is a literal `${ca-path}` reaching `load_verify_file()`. `NSCPClient`'s `connection_data` requires the handler; I'd make this one required too and have `AMissingHandlerLeavesThePathsVerbatim` pass `nullptr` explicitly.
- **`check_mk_client.hpp:26-33`** — the new `expand` lambda is applied to `ca` only. `certificate`, `certificate key` and `dh` still reach OpenSSL verbatim on the CLI/REST path, so `certificate=${certificate-path}/x.pem` fails there. (`NSCPClient` expands certificate and key but not `dh`.) Pre-existing, but the mechanism to fix it is now right there.
- **`nscp_client.hpp:146`** — `handler->expand_path(ssl.ca_path)` is now unconditional where the two lines above it are guarded by `!empty()`. The ctor requires a handler so it can't be null today, but `CheckMKClient` guards and this doesn't; worth being consistent.

### 6. Release numbering

Per `CLAUDE.md`, `codacy/git-version` scans commit messages, and the repo merges with merge commits that don't carry the PR title. The single commit's subject is `fix: verify the peer by default on the NSCP and check_mk clients`, so this reads as a patch — while the PR title says `feature:` and the upgrade note carries 💥. A user-visible breaking default change wants at least `feature:` in the commit subject, or the release it lands in is numbered wrong.

---

Nothing here blocks: the defaults, the settings plumbing, the tests and the two notes are all correct as written. Items 1 and 2 are the ones I'd want before this ships, since they're what an operator meets on upgrade day.

---

